PR #3, opened 2019-04-27
Update to use _Host.cshtml in Server app.

StevenTCramer:


marin-bratanov:
Hi Steven,

The client-side apps still need the `wwwroot/index.html` file, the proposed change is for server-side apps, and they are documented in a separate article where this information (and a couple of other differences) are taken into account: [https://github.com/telerik/blazor-docs/blob/master/getting-started/server-blazor.md](https://github.com/telerik/blazor-docs/blob/master/getting-started/server-blazor.md). It is live at [https://docs.telerik.com/blazor/getting-started/server-blazor](https://docs.telerik.com/blazor/getting-started/server-blazor)

StevenTCramer:
@marin-bratanov there are three scenarios 
1. server-side (this is one you documented https://docs.telerik.com/blazor/getting-started/server-blazor)
2. standalone client-side (this is the one you documented)
3. Hosted client-side  (this is the one I documented)

I don't use the standalone much and didn't think about it but yes it needs the index.html but the hosted clientside does not.  So maybe need to add the hosted docs.

marin-bratanov:
Thanks for following up, Steven.

We aim to document the ASP.NET Core hosted client-side scenario (see step 4 [here](https://docs.telerik.com/blazor-ui/getting-started/client-blazor#add-the-telerik-components-to-your-project)).

On my end, the hosted client-side template still generates a Client project with its wwwroot/index.html file where those changes need to go. The only difference is that it comes down from an ASP.NET Core project and not as a static site.

Can you confirm you are using the VS 2019 Preview version and that you have the latest [Blazor templates](https://marketplace.visualstudio.com/items?itemName=aspnet.blazor) installed? I have to update my templates (and run the CLI command) after each VS upgrade, otherwise my templates get broken.

PR #4, opened 2019-04-30
Fix ref value

StevenTCramer:
The ref value wasn't pointing to an actual property.


marin-bratanov:
Thank you for noticing and for taking the time to fix this, Steve :)

PR #5, opened 2019-05-14
Removed `id` from css

EdCharbeneau:
This `id` was for demo purposes only.


marin-bratanov:
Oh, yes, this is correct. There are probably a few other places that I will shortly tweak and then I will merge this

PR #17, opened 2019-08-28
Update grouping.md

willdean:
Groping is really unacceptable in the modern workplace ;-)


marin-bratanov:
Absolutely, Will. Thank you for noticing and putting those PRs in place. They'll be live shortly.

PR #18, opened 2019-08-28
Update filtering.md

willdean:
Wrong enum used in FilterRow section


marin-bratanov:
Thanks, Will, silly of me to miss that :)

PR #22, opened 2019-10-10
Update parameter values TItem and TValue

rycornell:


marin-bratanov:
Thank you for noticing, Ryan. I've updated your Telerik points as a small "thank you"

PR #23, opened 2019-10-14
Fix link

nemesv:


marin-bratanov:
Thank you for noticing and fixing this, it will be live shortly.

PR #27, opened 2019-12-20
Fixed typo, "row emplate" to "row template"

scdehmlow:


marin-bratanov:
Thank you for noticing and fixing this :)

PR #33, opened 2020-03-12
Fix typo "IQueryable" rather than "IQueriable"

scdehmlow:


marin-bratanov:
That was embarrassing, thank you for noticing and fixing it :)

PR #37, opened 2020-03-19
Added missing property "Class"

EduVencovsky:


marin-bratanov:
Thanks, nicely spotted :)

PR #40, opened 2020-03-25
Frozen columns

svdimitr:


svdimitr:
Added a commit to showcase the requested changes as follows:

- Tweaked the example with dynamic lock/unlock to have statically locked ID and Command columns (first and last) and moved that example as first. Added a new gif to showcase that
- Added heading and links for both examples
- Updated the link to the demo in the See Also section. (fixed my own mistake)

PR #44, opened 2020-03-31
Pager docs

svdimitr:


components/pager/overview.md
```diff
@@ -0,0 +1,118 @@
...
+tags: telerik,blazor,pager,paging
+published: True
+position: 20
+---
+
+# Pager Overview
+
+The `Pager` component will enable you to add page navigation in your Blazor application. It follows the design patter of our components but as standalone one can be used outside of Grid.
+
+To use Telerik Pager component for Blazor:
```
marin-bratanov: These instructions need to be shorter, drop the "optional" things. Add instructions when and where to provide the new set of data, and add a loop to render some items (even if it is just a few `<div>` elements with the ID of the item). This is the most important explanation on how to use the component and it needs to offer this information.

components/pager/overview.md
```diff
@@ -0,0 +1,118 @@
...
+}
+````
+
+>caption The result from the code snippet above
+
+![basic configuration of the pager](images/pager-basic-configuration-screenshot.jpg)
+
+## Features
+* `Class` - The CSS class that will be rendered on the main wrapping element of the Pager.
+* `Total` - **int** - Represents the total count of items in the pager.
```
marin-bratanov: The types of the parameters must also be in code blocks, not bold.

components/pager/overview.md
```diff
@@ -0,0 +1,118 @@
...
+* `ButtonCount` - **int** - The number of pages to be visible. To take effect the `ButtonCount` must be **less** than the pages count (ButtonCount < Total / number of items on the page)
```
marin-bratanov: maybe this is the "maximum number of pages to be visible". Confirm with dev.

components/pager/overview.md
```diff
@@ -0,0 +1,118 @@
...
+* `Page` and `@bind-Page` - **int** - Represents the current page of the pager. Those parameters are respectively for one and two-way data binding. If no `Page` or `@bind-Page` are provided they will default to the first page (1).
+* `PageChanged` - Fires when a new page is selected (used in one-way data binding).
+* `PageSize` - **int** - The number of items to be presented on a pages.
+
+## Examples
```
marin-bratanov: The first example should be something similar, but also include a `.Skip().Take()` call from an AllData collection into an CurrentPageData collection that you can use for the loop.

A second example here can call an `async` method that will return data based on the given PageSize and Page parameters so you can simulate load-on-demand from a service with it. Title the example accordingly.

components/pager/overview.md
```diff
@@ -0,0 +1,118 @@
...
+
+>caption Observe the behavior of the Pager with two-way data binding
+
+````CSHTML
+@*This example showcases the usage of Page and PageChanged in conjunction*@
+
+<TelerikPager Total="TotalItems"
+              ButtonCount="ButtonCount"
+              PageSize="ItemsOnPage"
+              @bind-Page="CurrentPage">
```
marin-bratanov: For two-way binding the example should show not only how the component puts the page in an external variable, but also how changing the variable affects the pager. For example, bind it to a NumericTextbox as well so people can see how the pager reacts to that as well.

marin-bratanov:
There must be a separate Events article.

components/pager/overview.md
marin-bratanov: The Localization demo is a good template for this

components/pager/overview.md
svdimitr: Done

components/pager/overview.md
svdimitr: Added demo

components/pager/overview.md
svdimitr: Done

components/pager/overview.md
svdimitr: Done

PR #54, opened 2020-04-23
Range selection docs

svdimitr:


components/calendar/selection.md
```diff
@@ -17,11 +17,14 @@ The user can select one or mode dates depending on the Calendar configuration se
...
 * `Multiple`
+* `Range`
 
 You can pre-select a date in Single selection mode by setting the `Value` property of the calendar to the desired date.
 
 To pre-select dates in the Multiple selection mode, use the `SelectedDates` property which is of type `List<DateTime>`.
 
+In `Range` selection mode you can get the start and end dates of a selected range, by the user, through the `RangeStart` and `RangeEnd` parameters of type `DateTime`. You also get events `RangeStartChanged` and `RangeEndChanged`. You can read more about them and see an example in the [Events]({%slug components/calendar/events%}) article.
+
 ## Receive User Selection
```
marin-bratanov, line 28: I would expect a third item in the list for the range selection

marin-bratanov: I also expect an example for handling range selection

marin-bratanov: The disabled dates also deserve their own example, so perhaps a total of 4 is in order (single selection, multiple selection, range selection, disabled dates) and they can all have headings so they can be linked

components/calendar/events.md
```diff
@@ -76,6 +78,72 @@ When handling the `ViewChanged` event, you cannot use two-way binding for the `V
 
 >tip You are not required to provide an initial value to the `View` parameter. It will default to `CalendarView.Month`.
 
+## RangeStartChanged
+
+The `RangeStartChanged` fires every time the user selects a new starting date for the range of dates. The first click on a date in the Calendar will always be the starting date and the range can be consisted of dates only onwards.
```
marin-bratanov, line 83: On a calendar with range selection only, perhaps? If so, clarify this in the sentence

marin-bratanov: Also, you can select earlier dates, but they are considered a start of the range too

components/calendar/events.md
```diff
@@ -76,6 +78,72 @@ When handling the `ViewChanged` event, you cannot use two-way binding for the `V
...
+}
+
+@code {
+    public DateTime Date { get; set; } = new DateTime(2020, 3, 1);
+    public DateTime RangeStart { get; set; } = new DateTime(2020, 3, 14);
+    public DateTime RangeEnd { get; set; } = new DateTime(2020, 3, 18);
+    public List<DateTime> SelectedDates { get; set; } = new List<DateTime>();
+
+    public void StartChangeHandler(DateTime startDate)
+    {
```
marin-bratanov, line 125: Add comments about changing the view model or else the user action will be cancelled. See other similar events for inspiration

components/calendar/events.md
marin-bratanov: The `RangeEndChanged` fires immediately after the start event with a weird date. Sync with dev and if this is expected, document it

marin-bratanov: It fires with the `default` value, and that makes sense. See the range picker docs I made

components/calendar/selection.md
svdimitr: Added 4 separate examples for Disabled Dates, Single, Multiple & Range selection. Commit: f29afdc730abf62698b27246678231ccb8302838

components/calendar/events.md
svdimitr: What i meant by *consisted of dates only onward.* is that the dates will be from the StartRange to the EndRange so onward from Start if that makes sense.

The rest of the comments are done with: 6517076ff1c6a5890590a43c1e44753654ca32e0.
Fixed a typo in: 2e3165f3ef7a59cf09397f57bff82ee98318fe07

components/calendar/events.md
svdimitr: Done with: 6517076ff1c6a5890590a43c1e44753654ca32e0

Issue #57, opened 2020-04-24
Find-ability of customization options in Charts


svdimitr:
- [x] Improve the find-ability of customization options
- [x] Maybe add more explanations to the documentations

marin-bratanov:
Let's close this after #59 is done

marin-bratanov:
Done with #62

PR #59, opened 2020-04-27
Update Chart documentation

svdimitr:
This pull request will collect commits which enhance the Chart documentation. 

**Articles updated**:
* Column Chart

### DIFF
--- a/components/chart/overview.md
+++ b/components/chart/overview.md
@@ -59,7 +59,7 @@ Basic chart and common settings/elements
 	};
 
 	public List<object> simpleData = new List<object>() { 10, 2, 7, 5 };
-	
+
 	public string[] xAxisItems = new string[] { "Q1", "Q2", "Q3", "Q4" };
 }
 ````
@@ -68,8 +68,7 @@ Basic chart and common settings/elements
 
 ![](images/overview-chart.png)
 
->tip When configuring nested properties and child elements in your chart, the inner tags will contain their parent tag name and add specifics to its end. You can see an example of this with the `ChartSeries` > `ChartSeriesLabels` tags in the above example.
-
+@[template](/_contentTemplates/chart/link-to-basics.md#configurable-nested-chart-settings)
 
 >caption Component namespace and reference
 
--- a/components/chart/types/column.md
+++ b/components/chart/types/column.md
@@ -80,6 +80,44 @@ The color of a series is controlled through the `Color` property that can take a
 
 @[template](/_contentTemplates/chart/link-to-basics.md#configurable-nested-chart-settings)
 
+@[template](/_contentTemplates/chart/link-to-basics.md#configurable-nested-chart-settings-categorical)
+
+>caption Configuring Label Rotation, Skipping the rendering of every second label and adding borders and padding to the Labels.
+
+````CSHTML
+@* Skip rendering every second label and customize them to have borders, padding and rotation. *@
+
+<TelerikChart>
+    <ChartSeriesItems>
+        <ChartSeries Type="ChartSeriesType.Column" Name="Product 1" Data="@series1Data">
+        </ChartSeries>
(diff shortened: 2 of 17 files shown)


marin-bratanov:
Let's unify the content and showcase the general approach for all chart types, as listing individual features is not feasible, there are too many.

Also, this might close #57

components/chart/types/column.md
```diff
@@ -59,6 +59,17 @@ Column series
 
 ## Column Chart Specific Appearance Settings
 
+### Markers
+
+Each data item is denoted with a marker (label). You can control and customize them through the `< ChartCategoryAxisLabels />` and its children tags.
```
marin-bratanov, line 64: The wording should be something like "Each item can have a label", because the markers are something different in the context of line-type series

marin-bratanov: Also, there is an extra space after the `<` bracket

marin-bratanov: The heading should be something like `Series Labels`. not markers

components/chart/types/column.md
```diff
@@ -59,6 +59,17 @@ Column series
...
+
+* `Visible` - hide all markers by setting this parameter to `false`.
+* `Step` - renders every n-th marker, where n is the value(double number) passed to the parameter.
+* `Skip` - skips the first n markers, where n is the value(double number) passed to the parameter.
+* `Angle` - rotates the markers with the desired angle by n degrees, where n is the value passed to the parameter. It can take positive and negative numbers. To set this parameter use the `< ChartCategoryAxisLabelsRotation />` child tag.
+
+To rotate the markers use the `ChartCategoryAxisLabelsRotation` child tag and set its `Angle` parameter. It can take positive and negative numbers as value.
```
marin-bratanov, line 71: Isn't this in the next template about configurable nested properties?

marin-bratanov: If so, maybe move this new section after the Color section so they are grouped together

components/chart/types/column.md
```diff
@@ -67,6 +78,47 @@ The color of a series is controlled through the `Color` property that can take a
...
+    <ChartTitle Text="Quarterly revenue per product"></ChartTitle>
+
+    <ChartLegend Position="ChartLegendPosition.Right">
+    </ChartLegend>
+</TelerikChart>
+
+@code {
+    public List<object> series1Data = new List<object>() { 10, 2, 5, 6, 8, 8, 13, 11, 4, 9, 10, 15, 14, 3, 2, 7 };
+    public List<object> series2Data = new List<object>() { 5, 8, 2, 7, 6, 11, 14, 13, 8, 7, 2, 7, 5, 9, 11, 10 };
+    public string[] xAxisItems = new string[] { "Q1", "Q2", "Q3", "Q4", "loooooooong label", "loooooooong label", "loooooooong label", "loooooooong label", "loooooooong label", "loooooooong label", "loooooooong label", "loooooooong label", "loooooooong label", "loooooooong label", "loooooooong label", "loooooooong label" };
```
marin-bratanov, line 117: Create data in a loop, you could use sequential numbers like "looong label N" so it is clearly visible what's skipped and rendered.

components/chart/overview.md
```diff
@@ -68,8 +68,7 @@ Basic chart and common settings/elements
 
 ![](images/overview-chart.png)
 
->tip When configuring nested properties and child elements in your chart, the inner tags will contain their parent tag name and add specifics to its end. You can see an example of this with the `ChartSeries` > `ChartSeriesLabels` tags in the above example.
-
+@[template](/_contentTemplates/chart/link-to-basics.md#configurable-nested-chart-settings)
```
marin-bratanov, line 71: I liked the original green tip that told people to look for nested tags. Perhaps add it back (or a variation of it) at the end of the `### Configurable Nested Chart Settings ` section

marin-bratanov:
History messed up, will be done with https://github.com/telerik/blazor-docs/pull/62

PR #78, opened 2020-06-05
Spelling fix.

hensleyrob:


marin-bratanov:
Thank you for noticing and fixing this, Rob!

PR #80, opened 2020-06-10
Updates to the Chart Tooltip Docs

svdimitr:


_contentTemplates/chart/link-to-basics.md
```diff
@@ -186,7 +186,8 @@ This approach is not limited only to the Labels - it can be used with to all tag
 #end
 
 #context-parameter-information
-* `Value` - maps to the value of the data point that is being hovered.
+* `FormattedValue` - maps to the value of the data point that is being hovered, formatted as a string.
```
marin-bratanov, line 189: Maybe add a sentence here that it contains the default rendering of the tooltip, rather than the value. For example, in more complex charts (such as OHLC) it will probably be an actual `<table>`.

components/chart/tooltip/shared.md
```diff
@@ -101,23 +101,25 @@ Each `Point` contains the following data:
...
-           }
+            @{
+                var points = context.Points;
+
+                foreach (var point in points)
+                {
+                    <div>
+                        <TelerikIcon Icon="@IconName.Information" />
+                        @*this example shows how to parse the FormattedValue to a int,*@
+                        @*but you can also parse it to any other type*@
```
marin-bratanov, line 117: maybe should be something like "any other type your data uses/has"

PR #87, opened 2020-06-19
Added documentation for Row click and double click events

svdimitr:


components/grid/events.md
```diff
@@ -42,31 +44,133 @@ The command buttons of a grid provide an `OnClick` event before firing their bui
...
 
-### PageChanged
+### OnRowClick
 
-The event fires when the user pages the grid.
+The `OnRowClick` event fires as a response to the user clicking on a row of the Grid. Clicking on the `GridCommandButton`, `CheckBox`, expanding a `Detail Template` or when the row is in `edit mode` will not trigger the event.
+
+It exposes a `GridRowClickEventArgs` object which provides data for the currently clicked row.
+
+>caption Use the OnRowClick event to receive information on the clicked row
```
marin-bratanov, line 53: I want this example to load data on demand - like  the chart one, and like this one https://docs.telerik.com/blazor-ui/components/grid/selection/single#selecteditemschanged-event - of course, `async`, and the selection article should mention that if you want to load data on demand you should use the click event, while selection is suitable for using data that's already in the current model - maybe even touch the selection example to just console.writeline something from the model and remove the nested grid to avoid confusion.

Maybe then remove the gif from the example if it is too large or too slow.

components/grid/events.md
```diff
@@ -42,31 +44,133 @@ The command buttons of a grid provide an `OnClick` event before firing their bui
...
+        public string Team { get; set; }
+        public DateTime HireDate { get; set; }
+    }
+}
+
+````
+
+>caption The result from the code snippet above
+
+![onrowclick basic example](images/onrowdoubleclick-example.gif)
```
marin-bratanov: remove this image, as it is not sufficiently clear and would not bring much value.

components/grid/events.md
```diff
@@ -42,31 +44,133 @@ The command buttons of a grid provide an `OnClick` event before firing their bui
 
 Fires when the item selection is enabled and the user changes the selected [item]({%slug components/grid/selection/single%}#selecteditemschanged-event) or [items]({%slug components/grid/selection/multiple%}#selecteditemschanged-event).
 
-### PageChanged
```
marin-bratanov, line 45: Please restore the PageChanged event docs, the easiest way to get them is to copy their section from the `master` branch

svdimitr: Done

PR #88, opened 2020-06-19
docs(dropdownlist): add defaulttext option for model binding

edrohler:
Added option for setting the default value to allow end-users to select a null value like, 0, or null.

### DIFF
--- a/components/dropdownlist/data-bind.md
+++ b/components/dropdownlist/data-bind.md
@@ -51,14 +51,15 @@ To bind the DropDownList to a model:
 
 1. populate its `Data` property with the collection of items you want in the dropdown
 1. set the `TextField` and `ValueField` properties to point to the corresponding names of the model
-1. set the `Value` property to the intial value of the model. If not set, it will be populated with the first item in the data source.
+2. set the `Value` property to the initial value of the model. If not set, it will be populated with the first item in the data source or the `DefaultText`.
+3. (optional) set `DefaultText` value. This will show when the selected value isn't in the collection of items. For example, 0 for int and null for an int? or string.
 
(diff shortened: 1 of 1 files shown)


components/dropdownlist/data-bind.md
```diff
@@ -51,14 +51,15 @@ To bind the DropDownList to a model:
 
 1. populate its `Data` property with the collection of items you want in the dropdown
 1. set the `TextField` and `ValueField` properties to point to the corresponding names of the model
-1. set the `Value` property to the intial value of the model. If not set, it will be populated with the first item in the data source.
+2. set the `Value` property to the initial value of the model. If not set, it will be populated with the first item in the data source or the `DefaultText`.
```
marin-bratanov, line 54: The Value will not be populated with the DefaultText, that's just an appearance feature. The Value can be null or the default value of the type. Please also check the Considerations section and let me know if you still find the docs lacking https://docs.telerik.com/blazor-ui/components/dropdownlist/data-bind#considerations

components/dropdownlist/data-bind.md
```diff
@@ -51,14 +51,15 @@ To bind the DropDownList to a model:
...
+3. (optional) set `DefaultText` value. This will show when the selected value isn't in the collection of items. For example, 0 for int and null for an int? or string.
```
marin-bratanov, line 55: I'd rather not burden the user with many steps

marin-bratanov:
Linked the sections that explain that from the top of the article in https://github.com/telerik/blazor-docs/commit/1cbe98e2e5e2adf78402c7d1bfc36be937c7e00e

PR #103, opened 2020-07-17
Update the Data Binding article for the Grid

svdimitr:


components/grid/columns/bound.md
```diff
@@ -84,10 +84,14 @@ You can use the following properties on the bound columns:
 * `Reorderable` - whether the user can [drag to reorder]({%slug components/grid/columns/reorder%}) this column.
 * `Resizable` - whether the user can [resize]({%slug components/grid/columns/resize%}) this column.
 * `Width` - the width of the column. See the [Dimensions]({%slug common-features/dimensions%}) article. Keep in mind that these are columns in a `<table>`, so it is often a good practice to leave one column without explicit dimensions so it can accommodate the remaining width and changes to the container size.
+* `AutoGenerateColumns` - this parameter allows the Grid to [automatically generate columns]({%slug grid-columns-automatically-generated%}) based on the public properties of the model.
```
marin-bratanov, line 87: Is this on the column level?

svdimitr: No, this is on the Grid level. <Grid Data=Data AutoGeneratedColumns=true>

marin-bratanov: Then, it does not belong here as a parameter list for the column. Perhaps mentioning automatic column generation can be a note or a tip similar to the several other tips or notes.

svdimitr: On column level we have `ColumnWidth` which is only applicable to `<GridAutoGeneratedColumns >`, should I add it instead

marin-bratanov: `ColumnWidth` is not on the column level, it is in the `<AutoGeneratedColumns />` tag which is not in the scope of this article.

svdimitr: Updated here 536942e4dfa887c9da9dec315e533d1e0b5aabff

PR #127, opened 2020-09-01
Add docs for visible columns in TreeList

svdimitr:


marin-bratanov (changes requested):
Use shorter (generated) data (like from any other article but Overview), redo the screenshots at 100% DPI.

marin-bratanov (approved):

PR #156, opened 2020-10-21
Fix the snippet for adding TelerikRootComponent

jakzale:
When following the [First Steps with Server-Side UI for Blazor](https://docs.telerik.com/blazor-ui/getting-started/server-blazor) tutorial and applying it to default Blazor on .NET 5.0 RC2, I found out that the snippet for adding `<TelerikRootComponent>` will break the layout of the page.

`MainLayout.cs` needs the top-level `div` to have class `page`, otherwise the page content will be displayed below the menu (which only affects the desktop view of the page).

### DIFF
--- a/_contentTemplates/common/get-started.md
+++ b/_contentTemplates/common/get-started.md
@@ -177,21 +177,21 @@ Open the main layout file (by default, the `~/Shared/MainLayout.razor` file in t
         @inherits LayoutComponentBase
         
         <TelerikRootComponent>
-        
-            <div class="sidebar">
-                <NavMenu />
+            <div class="page">
+              <div class="sidebar">
+                  <NavMenu />
+              </div>
+          
+              <div class="main">
+                  <div class="top-row px-4">
+                      <a href="https://docs.microsoft.com/en-us/aspnet/" target="_blank">About</a>
+                  </div>
+          
+                  <div class="content px-4">
+                      @Body
+                  </div>
+              </div>
             </div>
-        
-            <div class="main">
-                <div class="top-row px-4">
-                    <a href="https://docs.microsoft.com/en-us/aspnet/" target="_blank">About</a>
-                </div>
-        
-                <div class="content px-4">
-                    @Body
-                </div>
-            </div>
-        
         </TelerikRootComponent>
 #end
 


marin-bratanov:
Thank you for this PR @jakzale we will probably add this for the release that provides .NET 5 compatibility. The tricky part is that adding this div breaks the layout for .NET 3.1.x projects, which is also bad considering that's the LTS release.

marin-bratanov:
So, it looks like the MS templates won't be backwards compatible, so this is the best explanation I could come up with https://github.com/telerik/blazor-docs/commit/32947502122cbfa2d9d5440e88b9b722362f5022#diff-e7ac565b0c8312e8706393df590a79d77f9d7ecdb6ed4860db1138435a25c102R181-R183

Do you think it's clear enough?

I also used that commit to update some old code that had fallen through the cracks.

It was easier for me to test things out in my local branch than to request permissions for your repo and fork it, which is why I will later decline this PR, but you will still find your Telerik points updated for bringing this up.

marin-bratanov:
I've actually separated the instructions into separate files like our [Convert wizard](https://docs.telerik.com/blazor-ui/getting-started/vs-integration/convert-project-wizard) does, and I will sync the New Project templates with that https://github.com/telerik/blazor-docs/commit/8bd8ff33940a704cd131039f90841f28b8adb460

Thank you again for bringing this up.

jakzale:
@marin-bratanov thanks!

I like the approach with having a separate `TelerikLayout` layout.

Given creating `TelerikLayout.razor` will be the first thing that most developers will do when trying out the library; in order to make that experience smoother, may I suggest:
- either bundle the layout with the library, using a descriptive name, such as `TelerikDefaultLayout`, or
- provide a template for `dotnet` cli?

marin-bratanov:
The idea about adding a layout in our package sounds pretty nice to me, let's see if it holds water with the rest of the team. In fact, I've already made a pull request that's pending a review.

On a CLI template, I see some issues with that which is why I've never considered it a really viable option:
* Making it generally available (in, say, nuget.org) will cause confusion as to how to get the actual packages that are on our private feed, with them being commercial. If the package with those templates is not public, it will be hard to get/find.
* A non-public package means you need to already know about it, so you have read our docs. At that point, it becomes the same as the wizards, but with extra steps.
* We can't possibly compete with the features that `dotnet new` can provide, and a new project often needs to scaffold authentication, databases, controllers and whatnot. Projects for our components are usually learning material to get started with them on the side, or to build a quick proof-of-concept. I think this will limit the usability of such templates a lot.
* Maintenance issues - yet another channel and tool to document in an already complex set of instructions, that already offers a lot of options that are hard to grasp for a newcomer. People preferring the CLI are rarely newcomers to the framework.

Am I missing or misunderstanding something on the benefits of a cli template and how it is to be used and made available?

PR #159, opened 2020-10-27
Update change-theme-runtime.md

uteschj:
Fixed typo.


marin-bratanov:
Good catch, thanks for the PR, your Telerik points were updated :)

PR #160, opened 2020-11-02
KB article: How to localize numeric labels for the Chart

svdimitr:


knowledge-base/chart-localized-labels.md
```diff
@@ -0,0 +1,114 @@
...
+I am having a Chart and I would like my labels to be localization aware.
+
+
+## Solution
+
+If you require your label to be a number only, you can use the [`Format`]({%slug components/chart/label-template-format%}#format-strings) parameter of the `<ChartSeriesLabels>` tag. It takes a [standard numeric format strings](https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings) and is localization aware.
+
+If you want to expose more information beyond a number, for example, some text, you can utilize the [`Template`]({%slug components/chart/label-template-format%}#templates) parameter, which is exposed from the `<ChartSeriesLabels>` tag. In a JavaScript file implement the desired number formatting by using the `toLocaleString` JavaScript method. An exemplary implementation can be seen in the example below:
+
+### Step by step explanation
```
marin-bratanov, line 34: I recommend putting each solution in its own section and list the sections before that, something like https://docs.telerik.com/blazor-ui/knowledge-base/grid-force-refresh

knowledge-base/chart-localized-labels.md
```diff
@@ -0,0 +1,114 @@
...
+
+1. Use a custom [template]({%slug components/chart/label-template-format%}#templates)
+
+2. Implement the desired number formatting function in a JavaScript file (in the example below, we will call it `template-helpers.js` and it resides in the `wwwroot` folder).
```
marin-bratanov, line 38: Add a hint here for the method that will be used, don't keep it just in the paragraph above, a lot of people go to steps/snippets immediately

knowledge-base/chart-localized-labels.md
```diff
@@ -0,0 +1,114 @@
...
+````JavaScript
+function formatNumberLocale(number, locale, decimals){
+    return number.toLocaleString(locale, { minimumFractionDigits: decimals, maximumFractionDigits: decimals });
+}
+````
+````Index
+<head>
+    <!-- there may be other content here -->
+    
+	<script src="~/template-helpers.js"></script>
```
marin-bratanov, line 105: Remove the `~/` as it may cause issues in a WASM app, the `<base>` tag will take care of the path

knowledge-base/chart-localized-labels.md
```diff
@@ -0,0 +1,114 @@
...
+        </ChartSeries>
+    </ChartSeriesItems>
+
+</TelerikChart>
+
+@code {
+    public class Data
+    {
+        public string XValue { get; set; }
+
```
marin-bratanov, line 76: You can remove these new lines. Consider putting the code for the template first, and maybe refactoring it into  property, something like `LocalizedTemplateString => GetFormattedNumericValue()` so the chart markup is cleaner

marin-bratanov (changes requested):

globalization/overview.md
```diff
@@ -25,7 +25,7 @@ The Telerik Components use the current thread culture to render the appropriate
 
 * [Calendar](https://demos.telerik.com/blazor-ui/calendar/globalization) - the names of the months and days of the week are taken from the current culture, and the `FirstDayOfWeek` of the culture is honored when ordering the days of the week. The Calendar uses the  `ShortestDayNames` array to get the short names for the days of the week and it expects them in the default order coming from the framework - Sunday to Saturday.
 
-* [Chart](https://demos.telerik.com/blazor-ui/chart/globalization) - the [Label Format Strings]({%slug components/chart/label-template-format%}) and [Tooltip]({%slug chart-tooltip-overview%}) are culture aware (e.g., currency, dates).
+* [Chart](https://demos.telerik.com/blazor-ui/chart/globalization) - the [Label Format Strings]({%slug components/chart/label-template-format%}) and [Tooltip]({%slug chart-tooltip-overview%}) are culture aware (e.g., currency, dates). Examples are available in the [How to localize numeric labels in the Chart]({%slug chart-kb-localized-numeric-labels%}) knowledge base article.
```
marin-bratanov, line 28: Prefer active to passive voide, something like "You can find examples in the"

knowledge-base/chart-localized-labels.md
```diff
@@ -0,0 +1,123 @@
...
+## Solution
+
+There are two different approaches:
+
+* [Using the Format Parameter](#using-the-format-parameter)
+* [Using the Template Parameter](#using-the-template-parameter)
+
+### Using the Format Parameter
+
+If you require your label to be a number only, you can use the [`Format`]({%slug components/chart/label-template-format%}#format-strings) parameter of the `<ChartSeriesLabels>` tag. It takes a [standard numeric format strings](https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings) and is localization aware.
```
marin-bratanov, line 37: Test if this works only for numbers. I suspect something like `Format="lorem ipsum {0:C}"` will work too, so maybe we can reword to something like "if you need to use only the Value of the series item, and a single line of text, you can use the format"

knowledge-base/chart-localized-labels.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Using the Template Parameter
+
+If you want to expose more information beyond a number, for example, some text, you can utilize the [`Template`]({%slug components/chart/label-template-format%}#templates) parameter, which is exposed from the `<ChartSeriesLabels>` tag. In a JavaScript file implement the desired number formatting by using the `toLocaleString` JavaScript method. An exemplary implementation can be seen in the example below:
```
marin-bratanov, line 41: Consider mentioning more than one line too, it is often important.

marin-bratanov: Also, "a sample implementation", "exemplary" in this context has a different connotation.

marin-bratanov (changes requested):

knowledge-base/chart-localized-labels.md
svdimitr: > Also, "a sample implementation", "exemplary" in this context has a different connotation.

This is done, but I believe that the brief mention should be enough. In the step by step section, I have added a link to the MDN documentation for the `toLocaleString` method.

knowledge-base/chart-localized-labels.md
svdimitr: It indeed does, so I have done the wording as suggested and also reword a little the entry paragraph for the solution that uses the Template.

marin-bratanov (approved):

PR #168, opened 2020-11-16
Add docs for the drag and drop feature of the TreeView

svdimitr:
Points of possible further discussion:

I have made the examples in separate sections because:

- They were too long to be on a single page
- Maybe it is not a bad idea to showcase how an app should be structured to the users as we have more and more tickets regarding the general behavior of  the framework

Shall we leave them like that?

### DIFF
--- a/components/treeview/drag-drop.md
+++ b/components/treeview/drag-drop.md
@@ -60,7 +60,6 @@ The `OnDrop` event provides an object of type `TreeViewDropEventArgs` to its eve
 ````Component
 @inject TreeViewFlatDataService TreeViewFlatDataService
 
-    <TelerikButton OnClick="@(() => Draggable = !Draggable)">Toggle Draggable</TelerikButton>
     <TelerikTreeView Data="@FlatData"
                      Draggable="@Draggable"
                      OnDrop="@OnDrop">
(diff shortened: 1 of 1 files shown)


marin-bratanov (approved):
Consider adding a sentence at the beginning of the "between trees" example to indicate whose OnDrop fires when - does it fire for the source or destination tree.

Perhaps also try adding the code as subsequent code snippets and not as tabbed code to see if it is more readable. If you already tried and it isn't - I am completely fine with it that way.

Also, is there a need for the button that toggles Draggale for the Flat treeview example? If not, we can probably remove it.

EDIT: Maybe add some explanations around the Position enum - maybe add a sentence when introducing it that the user will see a hint where they are dropping (above,below,over) if there is one. If there isn't - ignore this.

svdimitr:
> Consider adding a sentence at the beginning of the "between trees" example to indicate whose OnDrop fires when - does it fire for the source or destination tree.
> 
> Perhaps also try adding the code as subsequent code snippets and not as tabbed code to see if it is more readable. If you already tried and it isn't - I am completely fine with it that way.
> 
> Also, is there a need for the button that toggles Draggale for the Flat treeview example? If not, we can probably remove it.
> 
> EDIT: Maybe add some explanations around the Position enum - maybe add a sentence when introducing it that the user will see a hint where they are dropping (above,below,over) if there is one. If there isn't - ignore this.

Thank you for the comments, I will take the following actions:

1. I will remove the button for the Draggable.
1. I tried adding the code as subsequent code snippets and it became something like 300 + lines, which I found hard to read. 
1. I will add more explanations as suggested and re-request review

marin-bratanov:
I do not need to see it for two sentences, go ahead and merge

PR #172, opened 2020-11-18
KB article on how to get the selected content from the editor

svdimitr:

### DIFF
--- a/knowledge-base/editor-get-selection.md
+++ b/knowledge-base/editor-get-selection.md
@@ -2,8 +2,8 @@
 title: Get the content of the Editor selected by the user
 description: How to get the content of the Editor selected by the user
 type: how-to
-page_title: Get the content of the Editor selected by the user
-slug: editor-kb-content-selection
+page_title: Get the user selection in the Editor
+slug: editor-kb-get-selection
 position:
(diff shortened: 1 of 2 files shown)


knowledge-base/editor-get-selection.md
```diff
@@ -0,0 +1,157 @@
...
+````JavaScript
+function getSelectedText() {
+    var editorIframe = document.querySelector(".k-editor iframe");
+    return editorIframe.contentDocument.getSelection().toString();
+}
+````
+
+## Notes
+
+The browser selection exists only in the browser. Sending it to the Blazor app will remove all its functionality and its context. For example, you no longer have access to its start and end, you cannot alter them, and it will be up to your C# code to determine where exactly that selection is in the content and how to modify the entire content to both produce the desired results, and remain valid HTML. With this in mind, consider applying changes directly with JavaScript to the selection object.
```
marin-bratanov, line 154: Sync with dev whether it is possible to get the ProseMirror instance from the editor in JS so they can invoke commands from it. This will make altering the selection considerably easier, doing this manually is extremely difficult.

marin-bratanov (approved):

PR #178, opened 2020-12-01
fix typo

mburleigh:


marin-bratanov:
Thank you for noticing and fixing this, Matt! It will be live with our next upload. You'll also find your Telerik points updated.

PR #187, opened 2020-12-15
Add kb on getting the index of a grid row

svdimitr:


knowledge-base/grid-get-index-of-grid-row.md
```diff
@@ -0,0 +1,202 @@
...
+    List<MainModel> salesTeamMembers { get; set; }
+    List<MainModel> currentSalesTeamMembers { get; set; } //implementation of OnRead
+    MainModel currentItem { get; set; } //need to save the selected item to determine its new index
+    public string Result { get; set; }
+
+    void OnRowClickHandler(GridRowClickEventArgs args)
+    {
+        var clickedRow = args.Item as MainModel;
+
+        currentItem = clickedRow;
```
marin-bratanov, line 158: add a method call here that will get the index immediately - you already have the current data because of onread, so you should move the logic here

marin-bratanov (changes requested):

marin-bratanov (approved):

PR #199, opened 2020-12-29
Docs general improvements

svdimitr:


svdimitr:
Some tasks:

* [ ] Go through all events article and make sure the ValueChanged event always updates the value

I will revise this comment with new tasks

PR #206, opened 2021-01-15
Add docs on TreeView checkboxes

svdimitr:


components/treeview/checkboxes/single.md
```diff
@@ -0,0 +1,419 @@
...
+    }
+}
+````
+
+>caption The result of the code snippet above
+
+![checking single node example](images/single-node-checking-two-way-binding-example.png)
+
+
+### Checking and Selecting a single node
```
marin-bratanov, line 291: Does this work for multiple mode? If yes, I recommend moving this example to the overview article (in its own section) and add an example with multiple checkboxes too.

marin-bratanov (approved):

components/treeview/checkboxes/single.md
svdimitr: It works, I have moved it to the overview article, you are correct it belongs there

PR #207, opened 2021-01-15
Remove IconName class mentions

marin-bratanov:


components/contextmenu/navigation.md
```diff
@@ -47,26 +47,26 @@ To use the Context Menu for navigating between pages:
...
                         Url = "/profile",
-                        Icon = IconName.User
+                        Icon = "user"
                     },
                     new MenuModel()
                     {
                         Text = "Language Settings",
                         Url = "/language",
-                        Icon = IconName.Table
+                        Icon = "globe"
```
ntacheva, line 69: Table or globe?

marin-bratanov: "globe", I've messed something up, please commit the fix

svdimitr: According to me, it should be Globe, it is more suitable for the `Language setting`

svdimitr: it is globe right now, so nothing should be further committed. Not so long ago I ran this specific example

ntacheva (approved):

svdimitr (approved):

PR #213, opened 2021-01-21
Dropdown default text

ntacheva:


components/dropdownlist/overview.md
```diff
@@ -116,41 +131,17 @@ The DropDownList provides the following features:
...
 
     protected List<int> MyIntList = new List<int>() { 1, 2, 3 };
 
+    //Current value is 0 (no item is sellected) which allows the DefaultText to be displayed.
     protected int MyIntItem { get; set; }
 }
 ````
 
 
-
```
marin-bratanov, line 128: Indeed, this should be removed. I don't even remember why I had to add it. Maybe we've had a bug at some point.

marin-bratanov (changes requested):

marin-bratanov (approved):

PR #251, opened 2021-03-04
chore(docs): some improvements as per feedback comments.

ntacheva:


knowledge-base/common-icon-names-list.md
```diff
@@ -0,0 +1,456 @@
...
+slug: common-kb-icon-names
+position: 
+tags: 
+ticketid: 
+res_type: kb
+---
+
+## Description
+
+Since the `IconName` class is marked as obsolete and will soon be removed, below you will find a valid list of icon names to be hardcoded.
```
marin-bratanov, line 15: Reword this to be future proof, something like "if you have used the obsolete `iconname` class, here is a list of icon names in its stead"

Also, tweak the titles and file names to reflect that. We don't want this to combat the main article for popularity, it is an upgrading resource that shold be obsolete in less than a year

marin-bratanov (changes requested):
All ok, some tweaks on the font icons list upgrade kb

knowledge-base/common-icon-names-list.md
```diff
@@ -12,7 +12,7 @@ res_type: kb
 
 ## Description
 
-Since the `IconName` class is marked as obsolete and will soon be removed, below you will find a valid list of icon names to be hardcoded.
+If you have used the obsolete `IconName` class, here is a list of icon names in its stead.
```
marin-bratanov, line 15: maybe use "icon names" as a link to the current article, or even expand with a sentence that one shoud use strings now and that they are available in this and that article

marin-bratanov (approved):

PR #255, opened 2021-03-11
docs(componens-with-observable-data)

ntacheva:
docs(multiselect-observable-data):draft version of the article


components/multiselect/observable-data.md
```diff
@@ -0,0 +1,73 @@
...
+published: True
+position: 25
+---
+
+# MultiSelect with Observable Data
+
+@[template](/_contentTemplates/common/observable-data-intro.md#observable-data-intro)
+
+
+>caption Bind the MultiSelect component to an [ObservableCollection]({%slug common-features-observable-data%}), so it can react to collection changes.
```
marin-bratanov, line 16: Maybe the link here is not necessary, the article would not provide more details, but can confuse people and send them to read other articles too early.

marin-bratanov:
I think that there should be two other sections:

* in general for all those components there should be a second section that shows how to create a new collection
* for the multiselect there needs to be a section on updating the Value parameter - it is also a collection but it does not support observable

So most articles can have two sections (with a tableof contents in the beginning), the multiselect will have 3. Others (maybe the treeview) mightalso have 3 or more, depending on what they support and what needs explaining.

_contentTemplates/common/observable-data.md
```diff
@@ -0,0 +1,25 @@
...
+#refresh-data
+In Blazor, the framework will fire the `OnParametersSet` event of a child component (which is how child components can react to outside changes) only when it can detect a change in the object it receives through the corresponding parameter (like `Data` for the data sources of Telerik components). This detection works as follows:
+
+  * For primitive types (such as numbers, strings), this happens when their value changes.
+
+  * For complex types (such as data collections like `List`, or any `IEnumerable`, and application-specific models/objects), this happens when the object reference changes.
+
+    Thus, you would usually need to create a `new` reference for the view-model field (such as `TreeViewData = new List<MyTreeViewItem>(theUpdatedDataCollection);`) when you want the component to update.
+
+You can find some more explanations and examples for the Grid component in the [Force a Grid to Refresh]({%slug grid-force-refresh%}) Knowledge Base article.
```
marin-bratanov, line 24: I think this should be only in the grid-specific article not the common template.

_contentTemplates/common/observable-data.md
```diff
@@ -0,0 +1,25 @@
+#intro
+The most common reason you would use an ObservableCollection is to make a component (like a grid, treeview, treelist, dropdown) change or react when you change that collection.
+
+When you want to refresh the component data source like that, there are two important framework behaviors you need to be aware of.
```
marin-bratanov, line 4: Consider adding those behaviors here, something like

When you want to refresh the component data source like that, there are two important framework behaviors you need to be aware of - when ObservableCollection instances fire events, and how to refresh the data of a component when it is not an observable collection.

components/autocomplete/refresh-data.md
```diff
@@ -0,0 +1,150 @@
...
+title: Refresh Data
+page_title: Refresh Data
+description: Refresh Autocomplete Data using Observable Data or creating a new Collection reference.
+slug: autocomplete-refresh-data
+tags: telerik,blazor,autocomplete,observable,data,new,collection
+published: True
+position: 15
+---
+
+# Refresh Data
```
marin-bratanov, line 11: Include the component name in the heading text, it will make it easier for people and google to distinguish one article from the other. Applies to all component-specific articles

marin-bratanov (approved):

_contentTemplates/common/observable-data.md
ntacheva: Added.

components/autocomplete/refresh-data.md
ntacheva: Added.

_contentTemplates/common/observable-data.md
ntacheva: Removed from the template, will add in the Grid specific article.

PR #258, opened 2021-03-15
docs(editor): paste cleanup initial

marin-bratanov:
## General tasks for a new feature (delete if not applicable)

- [x] Check links to and from demos, demos descriptions and metadata
- [x] Update relevant KB articles or sample projects that may now be easier or cover the scenario that is not built-in


jivanova (approved):
Great job. 

Thanks for adding the information about pasting text and images from Word! I totally forgot to include it as limitation in the spec.

PR #260, opened 2021-03-17
Spelling error in Grid docs

BiggT76:
Changed contructor to constructor.


marin-bratanov:
Thank you for noticing and fixing, Tyler! This will be live with our next upload.

PR #281, opened 2021-04-06
kb(combobox):added kb for combo not colsing on selection

ntacheva:


knowledge-base/combobox-not-closing.md
```diff
@@ -0,0 +1,92 @@
...
+    {
+        public int MyValueField { get; set; }
+        public string MyTextField { get; set; }
+    }
+}
+````
+
+## Solution
+
+To avoid the above described behavior you need to bind the `Value` parameter to a field of the same type as the `ValueField` is bound to.
```
marin-bratanov, line 63: Add a link here to the KB about getting a model from the dropowns, that need is why most people try to bind the value like that

marin-bratanov (changes requested):

marin-bratanov (approved):

PR #282, opened 2021-04-06
chore(treeview): added section for load on demand with one model

ntacheva:


components/treeview/data-binding/load-on-demand.md
```diff
@@ -18,9 +18,108 @@ You don't have to provide all the data the treeview will render at once - the ro
...
+@code {
+    public List<ProductItem> HierarchicalData { get; set; }
+
+    public class ProductItem
+    {
+        public string ProductName { get; set; }
+        public int Id { get; set; } //will be used to identify the node, not for rendering in this example
+        public List<ProductItem> SubProducts { get; set; }
+        public bool Expanded { get; set; }
+        public bool HasChildren { get; set; }
```
marin-bratanov, line 52: Consider defaulting this field to true in the model, so you don't have to initialize it in the sample data. I imagine that would be more common so you don't have to check whether the item has children when loading it.

marin-bratanov (approved):

PR #283, opened 2021-04-08
chore(grid): improvements on column events article per feedback - added check for column name in the example

ntacheva:


marin-bratanov (changes requested):
I propose that we keep the current example as the first example, and add the tip and the new code as a second example after the first one.

marin-bratanov (approved):

PR #296, opened 2021-04-24
Docs: Grid and TreeList row Drag and Drop

svdimitr:


components/grid/drag-drop.md
```diff
@@ -0,0 +1,309 @@
...
+        <GridColumn Field="@(nameof(SampleData.Name))" Title="Employee Name" Groupable="false" />
+        <GridColumn Field="@(nameof(SampleData.Team))" Title="Team" />
+        <GridColumn Field="@(nameof(SampleData.HireDate))" Title="Hire Date" />
+    </GridColumns>
+</TelerikGrid>
+
+@code {
+    private void OnRowDropHandler(GridRowDropEventArgs<SampleData> args)
+    {
+        MyData.Remove(args.Item);
```
marin-bratanov, line 97: I recommend you add a little bit of comments on what the code for data manipulations does and to denote that it is up to the app to implement this correctly as per its needs, even though it is mentioned in the text. Add it to every sample.

Also, consider disabling sorting and filtering because that will break the ordering logic which is rather naiive in this sample, consider linking the article about getting a row index to hint at people that the indexes are not so clear-cut. Also applies to all code snippets

marin-bratanov (changes requested):
Apply the changes I made in this commit to the treelist as well https://github.com/telerik/blazor-docs/pull/296/commits/409ea4e06c1c1f2008fba2673fcad5d1523e94dc

See the other comment about adding code comments, it is important for people to know where the grid ends and where the business logic begins.

marin-bratanov (approved):

marin-bratanov:
Add a mention (maybe in the Basics sections) that the feature is enabled through a dedicated dragging column the grid adds for you as a first column. That's important because you can't drag the entire row.

PR #312, opened 2021-05-05
kb(loader,loadercontainer):add loadercontainer in the main layout

xristianstefanov:


marin-bratanov (approved):
Make sur the URL matches any updates in the blazor-ui repo before merging

PR #331, opened 2021-05-18
chore(kb, grid):kb extentions for grid columns rendering

xristianstefanov:
- Added tip for conditional columns in a Grid
- Added additional description for later added columns in a Grid

### DIFF
--- a/knowledge-base/grid-column-in-custom-component-is-last.md
+++ b/knowledge-base/grid-column-in-custom-component-is-last.md
@@ -34,7 +34,7 @@ This behavior stems from the order in which the framework initializes components
 
 All parent components are initialized before their child components.
 
-On initialization, child components can't be rendered before the parent components. This is why the order in the actual code can't affect the order on initialization - the grid receives the `GridColumn` instances in the order in which their components are initialized. The columns that are added later in the component lifecycle will also be last.
+On initialization, child components can't be rendered before the parent components. The grid receives the `GridColumn` instances in the order in which their components are initialized. Components rendering determines the order of the columns initialization. Columns that are added programmatically later in the component lifecycle will also be last.
 
 You can observe and reproduce the same behavior with the example below.
 
--- a/knowledge-base/grid-column-in-custom-component-is-last.md
+++ b/knowledge-base/grid-column-in-custom-component-is-last.md
@@ -34,7 +34,7 @@ This behavior stems from the order in which the framework initializes components
 
 All parent components are initialized before their child components.
 
-On initialization, child components can't be rendered before the parent components. The grid receives the `GridColumn` instances in the order in which their components are initialized. Components rendering determines the order of the columns initialization. Columns that are added programmatically later in the component lifecycle will also be last.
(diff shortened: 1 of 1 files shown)


knowledge-base/grid-column-in-custom-component-is-last.md
```diff
@@ -34,7 +34,7 @@ This behavior stems from the order in which the framework initializes components
 
 All parent components are initialized before their child components.
 
-On initialization, child components can't be rendered before the parent components. This is why the order in the actual code can't affect the order on initialization - the grid receives the `GridColumn` instances in the order in which their components are initialized.
+On initialization, child components can't be rendered before the parent components. This is why the order in the actual code can't affect the order on initialization - the grid receives the `GridColumn` instances in the order in which their components are initialized. The columns that are added later in the component lifecycle will also be last.
```
dimodi, line 37: Should be "order **of** initialization":

```suggestion
On initialization, child components can't be rendered before the parent components. This is why the order in the actual code can't affect the order of initialization - the grid receives the `GridColumn` instances in the order in which their components are initialized. The columns that are added later in the component lifecycle will also be last.
```

Also, the following statement may raise the question "What determines the order of column initialization, if it isn't the order in the code?":

"the actual code can't affect the order of initialization - the grid receives the `GridColumn` instances in the order in which their components are initialized"

Finally, what does "added later in the component lifecycle` mean? Is it programmatic addition of columns or just "later" in the code?

knowledge-base/grid-column-in-custom-component-is-last.md
```diff
@@ -107,6 +107,7 @@ You can observe and reproduce the same behavior with the example below.
 }
 ````
 
+>tip If you want to have conditional columns, you can use the [`Visible`]({%slug grid-columns-visible%}) parameter. With a certain condition, for example, a ternary operator or a method that returns a bool.
```
dimodi, line 110: The second sentence follows Bulgarian grammar rules (no subject, no action), but that approach is not used in English.

```suggestion
>tip If you want to have conditional columns, you can use the [`Visible`]({%slug grid-columns-visible%}) parameter. The condition can be a ternary operator or a method that returns a bool.
```

knowledge-base/grid-column-in-custom-component-is-last.md
```diff
@@ -34,7 +34,7 @@ This behavior stems from the order in which the framework initializes components
 
 All parent components are initialized before their child components.
 
-On initialization, child components can't be rendered before the parent components. This is why the order in the actual code can't affect the order on initialization - the grid receives the `GridColumn` instances in the order in which their components are initialized.
+On initialization, child components can't be rendered before the parent components. The grid receives the `GridColumn` instances in the order in which their components are initialized. Components rendering determines the order of the columns initialization. Columns that are added programmatically later in the component lifecycle will also be last.
```
dimodi, line 37:
```suggestion
On initialization, child components are always rendered after their parent component. On the other hand, the grid receives the `GridColumn` instances in the order in which their components are initialized. This determines why columns in a custom component will be initialized and rendered last. Furthermore, columns that are added programmatically later in the component lifecycle will also be rendered last.
```

dimodi (approved):

PR #337, opened 2021-05-21
chore(grid):added note for editor template context casting

ntacheva:


marin-bratanov (changes requested):
I recommend a Notes section for this so we can add examples, this single line might be a bit difficult to grasp for readers.

marin-bratanov (changes requested):
There are two more notes in the intro near the ToC, move them to the Notes section

marin-bratanov (approved):

PR #341, opened 2021-05-25
chore(chart):date axis - added section for numerical charts

ntacheva:


marin-bratanov (changes requested):
The "advanced features" section should, I think, also apply to numerical charts as well. Test it and if so - pull it back to the higher level and add it to the ToC.

Reduce the screenshots size - preferably to <800px width (you can easily do this by resizing your browser then refreshing the page with the chart).

Add a sentence that the numeric axis considerations will apply to other intervals and base units as well, so that the explanations are little more generic. Consider adding the info on how that could lead to a missing label at the end of the x-axis.

Add a note that the BaseUnit and Type settings on the x-axis for a numeric chart will probably become an enum and link the enhancement request so that people are clear that using this feature at the moment can lead to changes down the road that we do not consider breaking  - it is not implemented yet.

marin-bratanov (approved):

PR #345, opened 2021-05-27
Add a section to explain the behavior of the date input

svdimitr:


marin-bratanov (approved):
I tried to add more exapmles and explanatiosn to hopefully make things even clearer for people. If the text now feels too heavy and diffiult to read, feel free to remove most of them.

PR #349, opened 2021-06-01
Edit a Grid row on click with custom popup form

svdimitr:
Done for ticket 1521974

### DIFF
--- a/knowledge-base/grid-edit-on-row-double-click.md
+++ b/knowledge-base/grid-edit-on-row-double-click.md
@@ -41,7 +41,7 @@ The Grid exposes two events that allows you to respond to the user clicking on i
              OnRowDoubleClick="@OnRowDoubleClickHandler">
     <GridColumns>
         <GridColumn Field="@(nameof(SampleData.Id))" Width="120px" />
-        <GridColumn Field="@(nameof(SampleData.Name))" Title="Employee Name" Groupable="false" />
+        <GridColumn Field="@(nameof(SampleData.Name))" Title="Employee Name" />
         <GridColumn Field="@(nameof(SampleData.Team))" Title="Team" />
         <GridColumn Field="@(nameof(SampleData.HireDate))" Title="Hire Date" />
     </GridColumns>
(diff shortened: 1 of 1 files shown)


knowledge-base/grid-edit-on-row-double-click.md
```diff
@@ -0,0 +1,109 @@
...
+        <GridColumn Field="@(nameof(SampleData.Name))" Title="Employee Name" Groupable="false" />
+        <GridColumn Field="@(nameof(SampleData.Team))" Title="Team" />
+        <GridColumn Field="@(nameof(SampleData.HireDate))" Title="Hire Date" />
+    </GridColumns>
+</TelerikGrid>
+
+<TelerikWindow @bind-Visible="@isInEdit" Modal="true">
+    <WindowTitle>
+        Edit record
+    </WindowTitle>
```
ntacheva, line 53: May be it is valid to include WindowActions>Close and have a CancelEdit event subscribed to it?

knowledge-base/grid-edit-on-row-double-click.md
```diff
@@ -0,0 +1,109 @@
...
+@* Click on a Grid row twice to see the custom edit form *@ 
+
+<TelerikGrid Data="@MyData"
+             Height="400px"
+             Width="700px"
+             Pageable="true"
+             OnRowDoubleClick="@OnRowDoubleClickHandler">
+    <GridColumns>
+        <GridColumn Field="@(nameof(SampleData.Id))" Width="120px" />
+        <GridColumn Field="@(nameof(SampleData.Name))" Title="Employee Name" Groupable="false" />
```
ntacheva, line 44: We can remove the Groupable = "false" as this Grid is not groupable. I suppose it is left from the Grid overview example if you got it from there.

knowledge-base/grid-edit-on-row-double-click.md
```diff
@@ -0,0 +1,109 @@
...
+			<td>Product</td>
+			<td>Grid for Blazor</td>
+		</tr>
+	</tbody>
+</table>
+
+
+## Description
+
+I would like to edit a row in the Grid when the user clicks or double-clicks on it, instead of using the command buttons.
```
ntacheva, line 26: May be include "popup form" in the description as well.

marin-bratanov: Does not have to be popup mode only, should work with inline too, and it is more commonly used with inline.

marin-bratanov (commented):
I highly recommend you add an examle of using the inline mode with the grid state. We have a demo for external edit form and a sample project for a custom popup form. While they can be linked as relevant resources, clicking a row to edit in inline is the most common case- adding some of the InCell speed to InLine. You could also link InCell mode for people looking for quick edits.

knowledge-base/grid-edit-on-row-double-click.md
svdimitr: The form should be closed by using a `Cancel` button or else depending on the need of the application. The correct way would be to close the popup by using a form button rather than a Window action button.

ntacheva: I would just rename the Form button to "Cancel" instead of "Clear". It looks to me more suitable for the current context.

ntacheva (approved):

PR #351, opened 2021-06-03
Grid column header template update

ntacheva:
- Changed the example not to contain a block element inside a span as it is an invalid html
- Removed the section for misplacing of the sort indicators as with the flex layout such an issue is not observed


marin-bratanov (approved):
The CSS was probably fixed internally in the Themes, maybe a couple of months ago, but was useful for >1.5 years.

The ID column used to showcase centering the title of the column, which is an extremely common request, I advise that you put back a solution. The problem is that it seems we have changed the rendering (maybe to fix other issues around filtering, see the CSS you could get rid of) and now using templates makes it impossible to center titles. Sync with dev, perhaps it is worth to push the header CSS class implementation.

ntacheva:
Indeed there have been a change in the rendering. I removed the div rom the ID column in the example as it is considered not a valid html to have a div inside a span. However, as centering the header content is really a very common request, I created this kb to cover this scenario https://docs.telerik.com/blazor-ui/knowledge-base/grid-wrap-and-center-column-text

marin-bratanov:
This can be merged and pushed live.

PR #353, opened 2021-06-08
KB: Open window from another component

ntacheva:


dimodi (approved):
How about `ticketid` and `tags` meta data?
Do you think the keywords "display" or "show" can be relevant to this scenario too? If customers use them, we should include them somewhere for SEO.

ntacheva:
Indeed, I forgot to add tags, thanks, added now. I think "Show" sounds relevant for the current context, included it as well.
TicketId is not explicitly set as there have been several tickets for such a scenario and the kb is to cover the general case.

PR #364, opened 2021-06-18
kb(grid):kb for null ref exception when filtering in Grid with data c…

xristianstefanov:
- Added kb for fixing Null Reference when filtering in Grid with data coming from a server


svdimitr (approved):

marin-bratanov:
This is not necessarily about server data, but more about anonymous data types such as `dynamic` or `ExpandoObject` or `DataTable`

PR #371, opened 2021-06-24
chore(form):clarification for the Enabled parameter of the FormIem

ntacheva:
Improvement based on the following customer feedback from 21 Jun 2021 8:07 PM:
"This statement is confusing: Enabled - bool - whether the editor is disabled. Defaults to true and the Form takes its value from the [Enabled] data annotation attribute. Do you mean the [Editable] data annotation attribute?"

### DIFF
--- a/components/form/formitems/overview.md
+++ b/components/form/formitems/overview.md
@@ -27,8 +27,9 @@ The `FormItem` tag exposes the following parameters which you can use to customi
 
 * `Id` - `string` - maps to the `id` HTML attribute of the `<input>` tag.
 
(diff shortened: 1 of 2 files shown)


marin-bratanov (dismissed):

marin-bratanov:
Make sure to check the data attributes section in the overview article to ensure both are correct and complete https://docs.telerik.com/blazor-ui/components/form/overview#data-annotation-attributes

marin-bratanov (approved):

ntacheva:
Synced with @jivanova to confirm the correct Data Annotation Attribute that controls the `Enabled` property of the Form item - it is [Editable], not [Enabled]. Fixed in both Form Overview and Form Items Overview.

PR #378, opened 2021-06-28
chore(grid):link editing/overview notes section

ntacheva:
Included a link to the notes section of the editing/overview article as per the note in 1525025

### DIFF
--- a/components/grid/templates/editor.md
+++ b/components/grid/templates/editor.md
@@ -12,7 +12,7 @@ position: 15
 
 The column's `EditTemplate` defines the inline template or component that will be rendered when the user is [editing]({%slug components/grid/overview%}#editing) the field. It is also used when inserting a new item.
 
-You can data bind components in it to the current context, which is an instance of the model the grid is bound to. You will need a global variable that is also an instance of the model to store those changes. The model the template receives is a copy of the original model, so that changes can be cancelled (the `Cancel` command).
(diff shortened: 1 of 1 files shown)


components/grid/templates/editor.md
```diff
@@ -12,7 +12,7 @@ position: 15
 
 The column's `EditTemplate` defines the inline template or component that will be rendered when the user is [editing]({%slug components/grid/overview%}#editing) the field. It is also used when inserting a new item.
 
-You can data bind components in it to the current context, which is an instance of the model the grid is bound to. You will need a global variable that is also an instance of the model to store those changes. The model the template receives is a copy of the original model, so that changes can be cancelled (the `Cancel` command).
+You can data bind components in it to the current context, which is an instance of the model the grid is bound to. You will need a global variable that is also an instance of the model to store those changes. The model the template receives is a copy of the original model, so that changes can be cancelled with the `Cancel` command (more details on how and when that copy is created you can find in the notes section of [editing/overview](../editing/overview#notes) article).
```
dimodi, line 15:
```suggestion
You can data bind components in the editor template to the current context, which is an instance of the model the grid is bound to. You will need a global variable that is also an instance of the model to store those changes. The template receives a copy of the original model, so that changes can be cancelled with the `Cancel` command. See the **Notes** section of the [Editing Overview]({%slug components/grid/editing/overview%}#notes) article for more details on how and when that copy is created.
```

dimodi (changes requested):
Prefer shorter and simpler sentences with straight-forward line of thought and no subordinate clauses.

By the way, do we use relative URLs? These will break if we restructure the documentation site. On the other hand, incorrect slugs will be reported during build.

dimodi (approved):

PR #405, opened 2021-07-16
Update overview.md

LanceMcCarthy:
Quick language update to clarify that SaveUrl is not an optional property


LanceMcCarthy:
The language in this sentence has confused customers, they read it as both SaveUrl and RemoveUrl are optional. This change makes it clear that only the RemoveUrl is optional.

marin-bratanov (approved):

PR #409, opened 2021-07-25
docs(grid,treelist):add section and kb for delete confirmation dialog

xristianstefanov:
- Added folder with articles for Delete Confirmation dialog in Grid and TreeList Editing


components/grid/editing/built-in-dialogs/overview.md
```diff
@@ -0,0 +1,28 @@
...
+previous_url: /grid/built-in-dialogs
+position: 0
+---
+
+# Grid Built-in Dialogs
+You can enable built-in dialogs in Grid to provide an extra step for the end-user before a Grid operation.
+
+The Grid component can use built-in dialogs for: 
+
+* [delete confirmation]({%slug grid-delete-confirmation%}) - displays a confirmation dialog when the user clicks the "**delete**" command button.
```
marin-bratanov, line 17: There is also the popup editing built-in dialog. Is there a serious reason for this article and its folder?

marin-bratanov (changes requested):
Same comments apply to the treelist.

I think that the dedicated section for the confirm dialogs is enough and the separate article is not needed, unless we plan to add many other built-in dialogs, which I do not find likely.

If you want to add a new article for delete confirmation, I recommend a KB rather than such a deep nesting.

components/grid/editing/built-in-dialogs/delete-confirmation.md
```diff
@@ -0,0 +1,104 @@
...
+page_title: Grid - Delete Confirmation Dialog
+description: Enable delete confirmation dialog in Grid for Blazor.
+slug: grid-delete-confirmation
+tags: telerik,blazor,grid,delete,confirmation,dialog
+published: True
+position: 3
+---
+
+# Delete Confirmation Dialog
+The delete confirmation dialog triggers before item deletion. You can enable it by setting the `ConfirmDelete` parameter to `true`.
```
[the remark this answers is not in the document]
xristianstefanov: Done

components/grid/editing/built-in-dialogs/overview.md
xristianstefanov: We decided to create a folder for such confirmation dialogs because there is a high chance of adding other confirmation dialogs in the future.

marin-bratanov (approved):

PR #411, opened 2021-07-29
kb(multiselect): select all checkbox

marin-bratanov:
Inspired by Ticket ID: 1529738

Please merge and upload live.


ntacheva:
The issue in this scenario is that the Select all checkbox does not update the selected items in the viewport while the popup is opened. This may introduce confusing UX. Let's keep this open until after the release, I will take a look if we can work it around. If not, let's close it and add this kb once the Refresh method for the popup is available.

jivanova:
LGTM

PR #443, opened 2021-08-16
docs: Custom Wizard buttons do not fire OnChange

dimodi:


components/wizard/events.md
```diff
@@ -23,8 +23,13 @@ The OnChange event is triggered on the current step and fires before the step ha
 * `TargetIndex` - contains the index of the targeted new Wizard step.
 * `IsCancelled` - specifies whether the event is canceled and the built-in action is prevented.
 
->caption Handle the `OnChange` event of the first and second steps. The result from the snippet below.
+The OnChange event **does not fire** when clicking on [custom Wizard buttons]({%slug wizard-structure-buttons%}#custom-buttons). However, the event **will always fire** when clicking on the [Wizard Stepper]({%slug wizard-structure-stepper%}). This means that you may need to do one of the following:
```
ntacheva, line 26:
```suggestion
When using [custom Wizard buttons]({%slug wizard-structure-buttons%}#custom-buttons) the OnChange event **does not fire** on their click. However, the event **will always fire** when clicking on the [Wizard Stepper]({%slug wizard-structure-stepper%}). This means that you may need to do one of the following:
```

components/wizard/events.md
```diff
@@ -23,8 +23,13 @@ The OnChange event is triggered on the current step and fires before the step ha
...
+* Execute business logic in both the step `OnChange` handler and the [button `OnClick` handler]({%slug wizard-structure-buttons%}#custom-buttons).
+* Call the step `OnChange` handlers from the button `OnClick` handlers (example below).
```
ntacheva, line 28:
```suggestion
* Call the step `OnChange` handlers from the button `OnClick` handlers (second example below).
```

ntacheva (approved):

PR #454, opened 2021-08-26
docs: Improve TelerikRootComponent information

dimodi:


_contentTemplates/common/get-started.md
```diff
@@ -175,8 +174,7 @@ To setup a local NuGet package source, so you can install the Telerik components
         </TelerikRootComponent>
         
 
-1. Open the main layout file (by default, the `~/Shared/MainLayout.razor` file in the Blazor project), and add `@layout TelerikLayout` as the *first line* in the file:
-
+1. Open the main layout file (by default, the `~/Shared/MainLayout.razor` file in the Blazor project), and add `@layout TelerikLayout` as the *first line* in the file. This will ensure that the `TelerikRootComponent` wraps all the content in `MainLayout`. Alternatively, the `TelerikRootComponent` can reside directly in `MainLayout`, but we place it in another file for better separation of concerns.
```
marin-bratanov, line 177: Not sure if this addition will make it clearer, or will add confusion. Early versions had similar long-winded explanations that gave people options, and that caused questions and concerns like "which is better" or "which should I do".

troubleshooting/general-issues.md
```diff
@@ -59,7 +59,7 @@ There are several common cases when such a mismatch occurs:
 
 * The position and size of the `<app>` element (or however the root component of your Blazor app is called) does not match the `<body>`.
 
-* There is more than one `<TelerikRootComponent>` in the app (for example, a certain Razor Component has its own) - there should be only one instance in the main layout.
+* There is more than one `<TelerikRootComponent>` in the app (for example, a certain Razor Component has its own). There should be only one instance - either in `MainLayout`, or in a separate layout file, which is referenced in `MainLayout`.
```
marin-bratanov, line 62: Not sure if this addition will make it clearer, or will add confusion. Early versions had similar long-winded explanations that gave people options, and that caused questions and concerns like "which is better" or "which should I do".

_contentTemplates/common/get-started.md
dimodi: The current docs also triggers questions, which is the reason for the update. I added what we recommend at both places, so novice devs should have a clear choice, while the rest should be savvy enough.

PR #464, opened 2021-09-07
Timeline View for Scheduler

svdimitr:


components/scheduler/views/timeline.md
```diff
@@ -0,0 +1,110 @@
+---
+title: Timeline
+page_title: Scheduler - Timeline View
+description: Day View in the Scheduler for Blazor.
```
marin-bratanov, line 4: wrong text in the description field

marin-bratanov (changes requested):

components/scheduler/views/timeline.md
svdimitr: Thank you, corrected it

marin-bratanov (approved):

PR #482, opened 2021-09-10
Add Grid right align header cells KB

dimodi:


marin-bratanov (approved):
Can we not combine the cases into one rule that both sets padding and right-align and add a comment that the padding is for the sorting and filtering icons, and if they don't use these features they can simply remove that line? I'd propose such a simple snippet as the first snippet and if we want to be exhaustive we can add the second CSS snippet with the many cases.

PR #494, opened 2021-09-13
Add Grid and TreeList interface binding docs

dimodi:
* Moved the Grid data binding info from Overview to a separate article.
* The Grid and TreeList interface-related stuff is currently documented differently, in the way, which made most sense to me (Interface section in the single data binding article for the Grid, separate Interface article for the TreeList).

### DIFF
--- a/components/grid/columns/bound.md
+++ b/components/grid/columns/bound.md
@@ -1,6 +1,6 @@
 ---
 title: Data Binding
-page_title: Grid - DataBound Column
+page_title: Data Bound Columns | Grid for Blazor
 description: Data binding and bound column properties in Grid for Blazor.
 slug: components/grid/columns/bound
 tags: telerik,blazor,grid,bound,column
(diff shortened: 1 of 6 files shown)


marin-bratanov:
I think the column data binding articles link the data binding section in overview, maybe that link should now go to the new dedicated article.

dimodi:
Hm, I assume you are talking about the Grid. I can't find a link from any Grid docs page to the Data Binding section in Overview. Can you point it out?

marin-bratanov:
I thought this one does https://docs.telerik.com/blazor-ui/components/grid/columns/bound but it seems I'm wrong. But it should :)

dimodi:
OK, I added a link. It looks a bit like a patch, but the whole article needs restructuring, which will be done after the release. For example, it should be split to "Overview" and "Bound columns".

components/grid/columns/bound.md
```diff
@@ -1,17 +1,19 @@
 ---
 title: Data Binding
-page_title: Grid - DataBound Column
+page_title: Data Bound Columns | Grid for Blazor
```
ntacheva, line 3: Shall we consider this a better directive for the page title? A more common used one amongst the documentation is Component - FooBar

components/treelist/data-binding/interface.md
```diff
@@ -0,0 +1,127 @@
...
+    List<IModel> TreeListData { get; set; }
+
+    protected override void OnInitialized()
+    {
+        var data = new List<IModel>();
+
+        data.Add(new Model1()
+        {
+            Id = 1,
+            ParentId = null,
```
ntacheva, line 74: Snippet not runnable - error with ParentId = null. The field in the model is non-nullable.

ntacheva (approved):

PR #500, opened 2021-09-14
Scheduler events

svdimitr:


components/scheduler/events.md
```diff
@@ -598,6 +601,225 @@ To implement appointment editing, the scheduler exposes the `OnCreate`, `OnDelet
 }
 ````
 
+## OnItemClick
+
+The `OnItemClick` event fires when the user clicks on an appointment in the Scheduler. 
+It provides a `SchedulerItemClickEventArgs` object to the event handler and you can get the `Item` and cast it to your own model. If you set the `ShouldRender` field to `true`, the component will re-render. This can be useful if you need to change its parameters or state during the event execution and especially if you need to execute `async` logic in the event handler.
```
dimodi, line 607: https://stackoverflow.com/questions/295104/what-is-the-difference-between-a-field-and-a-property

```suggestion
It provides a `SchedulerItemClickEventArgs` object to the event handler and you can get the `Item` property and cast it to your own model. If you set the `ShouldRender` property to `true`, the component will re-render. This can be useful if you need to change the Scheduler parameters or state during the event execution and especially if you need to execute `async` logic in the event handler.
```

components/scheduler/events.md
```diff
@@ -598,6 +601,225 @@ To implement appointment editing, the scheduler exposes the `OnCreate`, `OnDelet
...
+        public DateTime End { get; set; }
+        public bool IsAllDay { get; set; }
+    }
+}
+````
+
+## OnItemDoubleClick
+
+The `OnItemDoubleClick` event fires when the user double clicks on an appointment in the Scheduler. 
+It provides a `SchedulerItemDoubleClickEventArgs` object to the event handler and you can get the `Item` and cast it to your own model. If you set the `ShouldRender` field to `true`, the component will re-render. This can be useful if you need to change its parameters or state during the event execution and especially if you need to execute `async` logic in the event handler.
```
dimodi, line 680:
```suggestion
It provides a `SchedulerItemDoubleClickEventArgs` object to the event handler and you can get the `Item` property and cast it to your own model. If you set the `ShouldRender` property to `true`, the component will re-render. This can be useful if you need to change the Scheduler parameters or state during the event execution and especially if you need to execute `async` logic in the event handler.
```

components/scheduler/events.md
```diff
@@ -598,6 +601,225 @@ To implement appointment editing, the scheduler exposes the `OnCreate`, `OnDelet
...
+        public DateTime End { get; set; }
+        public bool IsAllDay { get; set; }
+    }
+}
+````
+
+## OnItemContextMenu
+
+The `OnItemContextMenu` event fires when the user right clicks on an appointment in the Scheduler. 
+It provides a `SchedulerItemContextMenuEventArgs` object to the event handler and you can get the `Item` and cast it to your own model. If you set the `ShouldRender` field to `true`, the component will re-render. This can be useful if you need to change its parameters or state during the event execution and especially if you need to execute `async` logic in the event handler.
```
dimodi, line 753:
```suggestion
It provides a `SchedulerItemContextMenuEventArgs` object to the event handler and you can get the `Item` property and cast it to your own model. If you set the `ShouldRender` property to `true`, the component will re-render. This can be useful if you need to change the Scheduler parameters or state during the event execution and especially if you need to execute `async` logic in the event handler.
```

dimodi:
The identical explanations bug me a bit. If a reader checks all three events, they may feel as if a robot has produced the description. Isn't it possible to somehow merge the information for the three events?

svdimitr:
> The identical explanations bug me a bit. If a reader checks all three events, they may feel as if a robot has produced the description. Isn't it possible to somehow merge the information for the three events?

We did that multiple times for the sake of different code snippets and docs brevity. If we decide to take this route we should update the other components' Events articles as well.

PR #502, opened 2021-09-14
Adding Figma UI Kits Article

yordan-mitev:
Adding the **Telerik UI Design Kits for Figma** article that is required for the Design Story.


yordan-mitev:
> @yordan-mitev I guess we need to configure some redirects for the moved articles?

Thanks for the reminder, this is a must do.

ntacheva (approved):

PR #526, opened 2021-10-13
Repl docs

ntacheva:


jivanova:
We could add information about where to submit feedback - in our feedback portal. I will give you link on Monday.

In addition, we should clarify thet the sharing snippets in VS works on a single file only - multiple selection is not allowed

PR #564, opened 2021-11-03
kb(slider):add kb for max parameter zero in slider

xristianstefanov:
- Added KB for Slider with 0 Max parameter

### DIFF
--- a/knowledge-base/slider-max-parameter-zero-error.md
+++ b/knowledge-base/slider-max-parameter-zero-error.md
@@ -22,7 +22,7 @@ res_type: kb
 
 
 ## Description
-I'm using the Telerik Slider, and I noticed that I can't set the Max to 0. When I set the Max parameter to zero, the component throws an error.
+I'm using the Telerik Slider, and I can't set Max to 0. When I set the Max parameter to zero, the component throws an error.
 
 ## Error Message
 >warning ArgumentException: Telerik.Blazor.Components.TelerikSlider`1[System.Int32] requires the Max parameter.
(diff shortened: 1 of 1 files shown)


knowledge-base/slider-max-parameter-zero-error.md
```diff
@@ -0,0 +1,53 @@
+---
+title: Requires the Max parameter error in Slider
+description: Max parameter of zero throws an error
```
dimodi, line 3: Blazor Slider requires a Max parameter and throws an error if zero is set.

knowledge-base/slider-max-parameter-zero-error.md
```diff
@@ -0,0 +1,53 @@
...
+		<tr>
+			<td>Product</td>
+			<td>Slider for Blazor</td>
+		</tr>
+	</tbody>
+</table>
+
+
+## Description
+I'm using the Telerik Slider, and I noticed that I can't set the Max to 0. When I set the Max parameter to zero, the component throws an error.
```
dimodi, line 25: "noticed" doesn't contribute to customer context, as a person will not search by this word.

Consider quoting the exact error message below too:

```suggestion
I'm using the Telerik Slider, and I can't set Max to 0. When I set the Max parameter to zero, the component throws an error.
```

knowledge-base/slider-max-parameter-zero-error.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Error Message
+>warning ArgumentException: Telerik.Blazor.Components.TelerikSlider`1[System.Int32] requires the Max parameter.
+
+## Cause\Possible Cause(s)
+Since the max parameter is a required parameter, we are checking if the user has provided value. That being said, if the user doesn't set any value for the Max parameter, the parameter will stay with its default value and throw an error. The default value for a non-nullable type (like int) is 0, and when you manually set 0 to the Max parameter, the component acts like there is no Max parameter declared and throws an error again.
```
dimodi, line 31:
```suggestion
The `Max` parameter is a required attribute and the component is checking if it is set. If the developer doesn't set any value for Max, the parameter value will remain the default value of the `int` type and throw an error. The default value for the non-nullable `int` type is 0, and when you explicitly set 0 to the Max parameter, the component acts like there is no Max parameter declared and still throws the error.
```

knowledge-base/slider-max-parameter-zero-error.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Solution
+You can avoid this exception by using a nullable type of the bound value.
```
dimodi, line 34:
```suggestion
You can avoid the exception by using a nullable type for the Slider `Value`.
```

dimodi (approved):

PR #566, opened 2021-11-05
docs: Add Hide Grid Scrollbar KB

dimodi:


knowledge-base/grid-hide-vertical-scrollbar.md
```diff
@@ -0,0 +1,86 @@
...
+    </GridColumns>
+</TelerikGrid>
+
+<style>
+    .grid-no-scroll .k-grid-header,
+    .grid-no-scroll .k-grid-footer {
+        padding-right: 0;
+    }
+    .grid-no-scroll .k-grid-content {
+        overflow-y: auto;
```
marin-bratanov, line 59: if set to `auto` and the `Height` of the grid does not accommodate all rows, you will still get a scrollbar in the content which will put the headers and footers out of sync with the data cells. Perhaps should be `hidden`, or he caveat explained.

dimodi: Indeed, an explanation is good. The idea here is that even if the developer messes up the Grid declaration by setting a Height, the user will still be able to see all rows. The other option is to set `overflow:visible`, but that will look uglier (if the problem occurs).

marin-bratanov (approved):

knowledge-base/grid-hide-vertical-scrollbar.md
```diff
@@ -0,0 +1,97 @@
...
+
+How do I hide the Grid vertical scrollbar? I don't set the Grid `Height`, so I prefer to completely remove the disabled vertical scrollbar with CSS.
+
+## Solution
+
+There are three required CSS styles to disable vertical scrolling and ensure proper cell alignment between the header, data and footer area of the Grid/TreeList:
+
+1. Remove the vertical scrollbar, which is visible by default.
+1. Remove the empty space above and below the vertical scrollbar. These empty spaces are in the header and footer area.
+1. Remove some right borders in the header and footer area, which are no longer necessary.
```
dimodi, line 34:
```suggestion
1. Remove two right borders in the header and footer area, which are no longer necessary. These borders normally separate the empty spaces from the header and footer cells.
```

PR #582, opened 2021-11-18
kb(window):added kb for chat popup

xristianstefanov:
- Added KB for implementing a chat popup with the modal window
- Responsive chat


marin-bratanov (commented):
I strongly advise against such a KB. A Chat component has a huge amount of moving parts and other functionality people expect and need, and this will not cover their needs. If people have the chat component to put in the window, the chances are that this component already has its own functionality to create its own popup.

xristianstefanov:
> I strongly advise against such a KB. A Chat component has a huge amount of moving parts and other functionality people expect and need, and this will not cover their needs. If people have the chat component to put in the window, the chances are that this component already has its own functionality to create its own popup.

The idea is to have an example for the clients that want to just minimize the modal window as a chat, not a real alternative to the chat component. The feature request for the Chat component https://feedback.telerik.com/blazor/1517233-conversational-ui-chat will stay on "Unplanned".

dimodi:
We can remove the chat-related keywords and focus on the minimize behavior of the Window only.

xristianstefanov:
> We can remove the chat-related keywords and focus on the minimize behavior of the Window only.

I agree here to focus on the minimize behavior.

knowledge-base/window-modal-minimize-popup.md
```diff
@@ -0,0 +1,95 @@
+---
+title: How to minimize a popup window to the bottom of the page?
+description: 
+type: how-to
+page_title: How to minimize a popup window to the bottom of the page?
+slug: window-modal-minimize-popup
+position: 
+tags: window,modal,chat,popup,collapse,minimize
```
dimodi, line 8:
```suggestion
tags: window,modal,popup,collapse,minimize
```

knowledge-base/window-modal-minimize-popup.md
```diff
@@ -0,0 +1,95 @@
...
+
+@code {
+    string Title = "My Responsive Popup";
+    bool isModalVisible { get; set; } = true;
+
+    private bool Small { get; set; }
+}
+
+<style>
+    .messanger {
```
dimodi, line 84:
```suggestion
    .minimized {
```

knowledge-base/window-modal-minimize-popup.md
```diff
@@ -0,0 +1,95 @@
...
+        <WindowAction Name="Close" />
+    </WindowActions>
+</TelerikWindow>
+
+@if (isModalVisible == false)
+{
+    @if (Small)
+    {
+        <br />
+        <div @onclick="@( _ => isModalVisible = true )" class="messanger"><strong>@Title[0]</strong></div>
```
dimodi, line 65:
```suggestion
        <div @onclick="@( _ => isModalVisible = true )" class="minimized"><strong>@Title[0]</strong></div>
```

dimodi:
It will be better to minimize the Window instance itself, instead of a custom <div>. Change the size and position attribute values.

The huge benefit will be that the user will be able to drag the minimized Window to a random location on the screen. We can hide/show the Window content with custom CSS.

dimodi (changes requested):
It will be better to minimize the Window instance itself, instead of showing a custom div. Change the Window size and position attribute values, remember the old values, and then restore them.

The huge benefit will be that the user will be able to drag the minimized Window to a random location on the screen. We can hide/show the Window content with custom CSS.

marin-bratanov:
The downside will be that with responsive layouts the minimized window will fly all over the place if the user resizes the browser.
Ideally, this is to be solved with something like position:fixed for the minimized window, but you lose draggingthen.
Even more ideally, a feature lets it minimize to a zone/container that you can define in your layout, but that's feasible in kendo and webforms where youmanipulate the domdirectly (see MinimizeZonID in Webforms which enables [this](https://demos.telerik.com/aspnet-ajax/window/examples/radwindowandmdi/defaultcs.aspx)). I hope you are starting to see why I am opposed to this article :) It covers too little, and is too complex already, to be usable in a real-world app.

dimodi (approved):

PR #601, opened 2021-11-24
KB: Charts - toggles series on legend click

ntacheva:


svdimitr:
I really like this KB example, great job, I think there would be some value to play with the CSS of the custom legend so it renders closer to the built-in legend (the dash in front, the font-family, font-size). Maybe you can play with the positioning a bit too, so it is more coehisive. 
Other than that I have no valuable input, great job again.

marin-bratanov (changes requested):
Edit the opener post in the feature request with a sample, do not make a KB for features.
https://feedback.telerik.com/blazor/1442813-show-hide-series-on-legend-click

ntacheva:
The feature request already contains the workaround (indeed I can touch it a bit with some CSS). I considered that we can include a kb for it as we cannot know whether the client will be searching. Thus we can also cover a scenario when they look up such feature in the docs, land on the kb and then find both the workaround and the public feature request. What do you think, shall I reopen it?

PR #648, opened 2021-12-20
Fix alphabetical order of properties

ntacheva:


svdimitr:
I never knew we followed alphabetical order when listing the features/attributes/parameters of the component. Let's not merge this as when we revamp the overview articles this would be out of order either way.

dimodi (approved):
On second thought, we may also experiment with sorting by importance (Value and DebounceDelay first)

PR #669, opened 2021-12-27
Update fiddler jam article to generate less traffic

jivanova:
It turned out that our documentation pages are ranked better than the Fiddler documentation. Thus, we are reducing our content and Fiddler information in the article.


jivanova:
@telerik/blazor-support I will merge the change to go live now in order to have more time to decrease our ranking, but please review the PR and let me know if you have any feedback.

PR #675, opened 2022-01-04
chore(textbox): Add example for programmatically changing the value

ntacheva:
Based on customer feedback - apparently some customers find it hard to achieve.


components/textbox/overview.md
```diff
@@ -64,6 +64,27 @@ for example: https://demos.telerik.com/blazor-ui/textbox/password
 }
 ````
 
+>caption Programmatically change the textbox value
```
dimodi, line 67:
```suggestion
>caption Programmatically change the TextBox value
```

dimodi (approved):

PR #685, opened 2022-01-05
kb(input): important update in the input keyboard events kb for the d…

xristianstefanov:
- Added note for the new DebounceDelay parameter in the KB for capturing input keyboard events. It is important because it can break many scenarios.

### DIFF
--- a/knowledge-base/inputs-handle-keyboard-events.md
+++ b/knowledge-base/inputs-handle-keyboard-events.md
@@ -32,7 +32,7 @@ The keyboard events bubble up the DOM, so you can attach event handlers for them
 
 If you are looking for an event when the user confirms an action, consider the built-in `OnChange` event which fires when the user presses `Enter` or blurs the input.
 
->important Optionally, you can disable the [**`DebounceDelay`**](https://docs.telerik.com/blazor-ui/components/textbox/overview#:~:text=DebounceDelay,int) parameter of the textbox by setting it to 0. The `DebounceDelay` default value is 150. It can fire the `onkeypress` event logic before the [**`ValueChanged`**](https://docs.telerik.com/blazor-ui/components/textbox/events#valuechanged) event of the input. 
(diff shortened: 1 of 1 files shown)


knowledge-base/inputs-handle-keyboard-events.md
```diff
@@ -32,13 +32,15 @@ The keyboard events bubble up the DOM, so you can attach event handlers for them
 
 If you are looking for an event when the user confirms an action, consider the built-in `OnChange` event which fires when the user presses `Enter` or blurs the input.
 
+>important Optionally, you can disable the [**`DebounceDelay`**](https://docs.telerik.com/blazor-ui/components/textbox/overview#:~:text=DebounceDelay,int) parameter of the textbox by setting it to 0. The `DebounceDelay` default value is 150. It can fire the `onkeypress` event logic before the [**`ValueChanged`**](https://docs.telerik.com/blazor-ui/components/textbox/events#valuechanged) event of the input.
```
dimodi, line 35: Use slugs instead of hard-coded URLs.

Parameters cannot be disabled - they are just set. Features can be disabled.

dimodi (changes requested):
Use slugs

dimodi (approved):

knowledge-base/inputs-handle-keyboard-events.md
```diff
@@ -32,13 +32,15 @@ The keyboard events bubble up the DOM, so you can attach event handlers for them
 
 If you are looking for an event when the user confirms an action, consider the built-in `OnChange` event which fires when the user presses `Enter` or blurs the input.
 
+>important Optionally, you can set the [**`DebounceDelay`**]({%slug components/textbox/overview%}#features) parameter of the textbox to 0. The `DebounceDelay` default value is 150. It can fire the `onkeypress` event logic before the [**`ValueChanged`**]({%slug components/textbox/events%}#valuechanged) event of the input.
```
dimodi, line 35: Maybe we should explain this even better - does the DebounceDelay value determine the order of events? What is "it" in "it can fire"? DebounceDelay surely does not fire `keypress`.

PR #688, opened 2022-01-07
KB: Time duration input for more than 24 hours

ntacheva:


jivanova (approved):
🚀  great job. I would not think of maskedtextbox use case.

PR #698, opened 2022-01-11
docs(upload):added upload api methods

giatsidis:
- Added the new upload methods in upload-overview
- Added an example that uses the methods 
- Added link for methods in the upload-events page


components/upload/overview.md
```diff
@@ -172,7 +172,60 @@ The <a href="https://www.telerik.com/blazor-ui/upload" target="_blank">Blazor Up
...
+* `ClearFiles` - Clears all files from the list, both uploaded and in queue.
+  
+* `UploadFiles` - Uploads all valid selected files. Fires the [OnUpload]({%slug upload-events%}#onupload) event.
+
+* `OpenFileSelectAsync` - Triggers the browser's file select dialog.
+
+>caption Get a reference to the Upload and use its methods. The example uses the same controller as the above demo.
+
+````CSHTML 
+@* This example showcases the use of the methods. *@
```
jivanova, line 188: We need to add the controller configuration to the snippet. Please, assure that the snippet is runnable in Telerik Repl. If it  is not and should not be - add the skip repl div (might need to sync with the support engineers how to do this, but if we can have it runnable, we should)

PR #708, opened 2022-01-17
docs(splitter): Docs for SizeChanged and CollapsedChanged events

ntacheva:


components/splitter/events.md
```diff
@@ -145,9 +145,94 @@ Try resizing any of the panes by dragging the splitbars
 }
 ````
 
+## SizeChanged
+
+`SizeChanged` is triggered when the `Size` parameter of the corresponding pane is changed.
```
svdimitr, line 150: The `eventName` is triggered

components/splitter/events.md
```diff
@@ -145,9 +145,94 @@ Try resizing any of the panes by dragging the splitbars
...
+        PaneSize = size;
+        Console.WriteLine("Pane 1 size was changed. Current size: " + PaneSize);
+    }
+}
+````
+
+
+## CollapsedChanged
+
+`CollapsedChanged` is triggered when the `Collapsed` parameter of the corresponding pane is changed.
```
svdimitr, line 194: The same review as in the `SizeChanged` event

svdimitr (changes requested):

svdimitr:
Apart from the two small changes requested, I think that we should make a new structure for the Splitter component. Maybe we should go for something similar to the Grid, where we have a dedicated folder for the Columns, here we might have it for the Pane.

components/splitter/events.md
ntacheva: Updated

components/splitter/events.md
ntacheva: Updated

svdimitr (approved):

PR #715, opened 2022-01-19
chore(grid): update filter row documentation about disabled clear filter

jivanova:
https://demos.telerik.com/blazor-ui-dev/grid/filter-row


jivanova:
merging after sync with Dimo

PR #740, opened 2022-01-25
Improvement for "Conditionally Hide Hierarchical Grid Expand Button" article

mp9007:
Setting the CSS class "no-children" was inverted.

I was having a "null" issue because the row didn't have childs. The error was style showing if I clicked on the cell after hidding the expand icon.
To fix this, I added a CSS class to set the visibility of the .k-hierarchy-cell to "hidden" so the click is not triggered when clicking in the cell.


ntacheva:
Hi @mp9007,

Thank you for your feedback! The CSS class "no-children" was indeed inverted, so we fixed that. As for hiding the cells that do not have expand icons - this could break the layout, so we consider a safer approach would be to stop the pointer events of these cells. I have already submitted a [PR with both of the changes](https://github.com/telerik/blazor-docs/pull/744), so I am hereby closing the current one.

PR #755, opened 2022-02-01
Missing component options articles

svdimitr:


dimodi (changes requested):
There is an extra "button" word in all articles.

On a side note, this initiative seems like a huge copy-paste overkill. We should have used a common article in the Themes section or at least content templates.

ntacheva (approved):

dimodi (approved):

PR #769, opened 2022-02-02
chore(fileselect):add note for size recommendation

xristianstefanov:
- Added explanation about the file size problem that might occur and what we recommend for fixing it.

### DIFF
--- a/components/fileselect/overview.md
+++ b/components/fileselect/overview.md
@@ -40,6 +40,8 @@ The Telerik FileSelect for Blazor provides a [Stream](https://docs.microsoft.com
 }
 ````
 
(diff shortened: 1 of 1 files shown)


components/fileselect/overview.md
```diff
@@ -40,6 +40,8 @@ The Telerik FileSelect for Blazor provides a [Stream](https://docs.microsoft.com
 }
 ````
 
+>important To avoid possible issues with the file size, we recommend modifying the [Max WebSocket Message Size](http://barwicktechnology.com/2020/03/23/signalr-modify-the-max-websocket-message-size/) used for **SignalR** in your VS project.
```
dimodi, line 43:
```suggestion
>important To avoid issues with large file size, increase the [Max WebSocket Message Size](http://barwicktechnology.com/2020/03/23/signalr-modify-the-max-websocket-message-size/) used for **SignalR** in your Visual Studio project.
```

dimodi (approved):

PR #795, opened 2022-02-23
Update Figma UI Kits Article

yordan-mitev:
This PR updates the existing Figma UI Kits article:
- Updates the page structure of the UI kit
- Updates the images
- Adds a section about InstaRelinker
- Updates the UI kit naming

cc @jivanova


dimodi (approved):
Is there a special reason to use `&mbdash;` instead of a plain hyphen?

yordan-mitev:
> Is there a special reason to use `&mbdash;` instead of a plain hyphen?

Yes, our Style Guide recommends [to use the em dash](https://testdocs.telerik.com/style-guide/punctuation#lists) in definition lists and where the sentence has a greater pause or where a clarification follows.
Our docs don't render an em dash when we use a plain hyphen.

dimodi:
OK, but shouldn't we put spaces around the mdash then? It looks weird to me like that.

yordan-mitev:
> OK, but shouldn't we put spaces around the mdash then? It looks weird to me like that.

The spaces around the em dash are another aspect of its usage and the Style Guide says that we never surround it with spaces. The same rule is embraced by the Google Style Guide, APA Style Guide, and many others.

PR #796, opened 2022-02-23
[3.1] chore(grid): update docs for filters in state

Stamo-Gochev:
Related to https://github.com/telerik/blazor/issues/3937

> Note: I checked the checkbox for creating a pull request to production, but not getting one (maybe requires the pr to master to be merged). Either way, let me know if this is incorrect according to the deadline for the documentation [release candidate](https://github.com/telerik/blazor/milestone/59)

### DIFF
--- a/_contentTemplates/grid/state.md
+++ b/_contentTemplates/grid/state.md
@@ -112,7 +112,7 @@
 #filter-menu-from-code
 @* This snippet shows how to set filtering state to the grid from your code
   Applies to the FilterMenu mode *@
-  
+
 @using Telerik.DataSource;
 
 <TelerikButton ThemeColor="primary" OnClick="@SetGridFilter">set filtering from code</TelerikButton>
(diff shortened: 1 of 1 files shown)


_contentTemplates/grid/state.md
```diff
@@ -176,9 +176,11 @@
 }
 #end
 
-
 #filter-menu-default-filters
->note When the Grid `FilterMode` is `FilterMenu`, the component will create default `FilterDecriptor`s in its state for **all columns**. This allows filter templates to work seamlessly and with little code. If you want to alter filters for a column, you must either modify the existing descriptor, or replace it. Simply adding an additional `FilterDescriptor` will not show up in the UI, because the Grid uses the first descriptor for the given field for the filtering UI. Another implication is that the Grid state always contains filter descriptors, no matter if the user has filtered or not. Inactive filter descriptors are distinguished by their `null` `Value`.
+
+>note Until version `3.1.0`, there was a bug that made the component create default `FilterDecriptor`s in its state for **all columns** when `FilterMode` is `FilterMenu`, including when no filtering occurs at all. This side-effect allowed filter templates to work seamlessly and with little code. After version `3.1.0`, ths bug was fixed and empty `FilterDecriptor`s are not added to the state of the component if no filtering occurs (the way filter templates work is unchanged).
```
dimodi, line 181: This sounds like filter templates no longer work seamlessly.

Do I understand correctly that now the state will include filter descriptors only for columns with applied filters?

If there is no breaking change and filter templates work in the exact same way, we can remove the "seamless" part.

Stamo-Gochev: > This sounds like filter templates no longer work seamlessly.

Not sure what do we mean by *seamlessly* - as explained in https://github.com/telerik/blazor/issues/3937, templates will continue to work in the same way, the change is that the state will not be filled with "empty filters", which is generally unnecessary if no filtering occurs at all. 

> 
> Do I understand correctly that now the state will include filter descriptors only for columns with applied filters?

Yes, but this part is also unchanged - even now, filtering a column adds a filter to the state with the selected filter criteria.

> If there is no breaking change and filter templates work in the exact same way, we can remove the "seamless" part.

It is rather a bug (a side-effect of a bug, incorrect implementation, etc.). If no interaction occurs, there is no good reason to add anything to the state - we rather do the opposite - when a user interacts with the component, their changes are reflected in the state.

svdimitr (approved):

PR #809, opened 2022-02-28
[3.1] docs: Add FloatingLabel documentation

dimodi:
Related to https://github.com/telerik/blazor/issues/3103

### DIFF
--- a/components/floatinglabel/overview.md
+++ b/components/floatinglabel/overview.md
@@ -28,7 +28,7 @@ The Telerik FloatingLabel enhances HTML `label` functionality in the following w
 1. Use the `TelerikFloatingLabel` tag.
 1. Set the `Text` parameter of the floating label.
 1. Place a [compatible Telerik component](#compatibility) inside the floating label.
-1. Set an `Id` to the Telerik component. Otherwise the floating label will not render a `for` attribute and label clicks will not focus the associated component.
+1. Set an `Id` to the Telerik component. The `Id` value will automatically render as a `for` attribute for the `<label>` element.
 
 >caption How to use a Floating Label
 
--- a/components/floatinglabel/overview.md
+++ b/components/floatinglabel/overview.md
@@ -1,0 +1,106 @@
+---
+title: Overview
+page_title: FloatLabel Overview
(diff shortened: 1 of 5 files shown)


components/floatinglabel/overview.md
```diff
@@ -0,0 +1,106 @@
...
+<style>
+    article style + table {
+        table-layout: auto;
+        word-break: normal;
+    }
+</style>
+
+| Attribute | Type | Description |
+| --- | --- | --- |
+| `Class` | `string` | Renders additional CSS class to the `span.k-floating-label-container` element, which holds the `<label>` child element. Use `Class` to apply custom styles or [override the theme]({%slug themes-override%}). |
```
Stamo-Gochev, line 94: Do we need to specify the exact rendering and classes here? It seems to me that explanation like "renders the class on the wrapper (top-most) element" works for most if not all components.

components/floatinglabel/overview.md
```diff
@@ -0,0 +1,106 @@
...
+| `Id` | `string` | Renders an `id` attribute to the  `label.k-label` element. To improve accessibility, set it together with the `AriaLabelledBy` parameter of the associated Telerik component. |
```
Stamo-Gochev, line 95: I would suggest something like "renders an id attribute to the label, so that it can match the `AriaLabelledBy` parameter of the associated Telerik component.

components/floatinglabel/validation.md
```diff
@@ -0,0 +1,83 @@
...
+description: The FloatingLabel can change its styling when it is inside a validated form and the associated form field is invalid. How to integrate the Blazor floating label with form validation.
+slug: floatinglabel-validation
+tags: telerik,blazor,floatinglabel,floating,label,form,validation
+published: True
+position: 10
+---
+
+# Floating Label Validation
+
+The Blazor FloatingLabel integrates with form validation. When a form field is invalid, the floating label will change color to suggest user action.
```
Stamo-Gochev, line 13:
```suggestion
The Blazor FloatingLabel integrates with form validation of compatible Telerik components. 
```
Although it is specified in the overview article, it might still be useful to have 3 more words here, so that there is less risk customers imply that this will work with any form component.

giatsidis (approved):

PR #810, opened 2022-02-28
[3.1] Grid, Treelist, Gantt - customize filtering docs

ntacheva:


common-features/filter-operators.md
```diff
@@ -0,0 +1,34 @@
...
+tags: telerik,blazor,filter,operator
+published: True
+position: 10
+---
+
+# Filter Operators
+
+The `FilterOperator` enum contains all filter operators supported in the Telerik UI for Blazor suite.
+
+When customizing the filtering behavior of the components, it is important to use the correct filter operators depending on the specific data type. Otherwise, you will receive an exception - the assigned `DefaultFilterOperator` not applicable for the current data type.
```
dimodi, line 15: Consider different text logic, which allows shorter sentences. For example something like this:

```suggestion
Each filter operator is compatible with a specific data type. Always use the correct filter operator during programmatic filter customizations. Otherwise, an exception will occur about the assigned `DefaultFilterOperator` not applicable for the current data type.
```

_contentTemplates/common/filtering.md
```diff
@@ -0,0 +1,12 @@
+#filter-row-customization-properties
+| Parameter      | Type | Description
+| ----------- | ----------- | -----------|
```
dimodi, line 3: Either use readable markdown, or minimalistic. Something in-between may look weird during future maintenance.

```suggestion
| Parameter | Type | Description |
| --- | --- | --- |
```

dimodi (approved):

PR #815, opened 2022-03-02
docs(gantt): fixed broken link

giatsidis:
- Fixed broken link in editor template page

### DIFF
--- a/components/gantt/gantt-tree/templates/editor.md
+++ b/components/gantt/gantt-tree/templates/editor.md
@@ -10,7 +10,7 @@ position: 15
 
 # Editor Template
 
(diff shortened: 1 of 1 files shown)


components/scheduler/editing/delete-confirmation-dialog.md
```diff
@@ -16,7 +16,7 @@ The delete confirmation dialog triggers before event deletion. You can enable it
 
 >caption Delete Confirmation dialog appearance
 
-![](images/scheduler-delete-confirmation.gif)
+![asdfsd](../images/scheduler-delete-confirmation.gif)
```
svdimitr, line 19: Why is `asdfsd` here? Use something more useful

svdimitr (changes requested):

components/scheduler/editing/delete-confirmation-dialog.md
svdimitr: You can use something like: `A gif that visualizes the delete confirmation`.

svdimitr (approved):

PR #862, opened 2022-03-29
[3.2] docs (selects) DebounceDelay and FilterDebounceDelay docs

ntacheva:


components/autocomplete/filter.md
```diff
@@ -18,6 +18,8 @@ Filtering ignores casing and the default filter operator is `starts with`. You c
 
 To control when the filter list appears, set the `MinLength` parameter. This can be useful if you have a very large list of data.
 
+By default, the filtering will be debounced with 150ms. You can configure that through the [`DebounceDelay`]({%slug autocomplete-overview%}#parameters) parameter of the component.
```
dimodi, line 21:
```suggestion
By default, the filtering will be debounced with 150ms. Configure that with the [`DebounceDelay`]({%slug autocomplete-overview%}#parameters) parameter of the component.
```

components/autocomplete/overview.md
```diff
@@ -86,6 +86,7 @@ The Blazor AutoComplete provides various parameters that allow you to configure
 | Parameter    | Type  | Description |
 | ----------- | ----------- | -------|
 | `Data` | `IEnumerable<TItem>` | allows you to provide the data source. Required.
+| `DebounceDelay` | `int` <br/> 150 | Time in milliseconds between the last typed symbol and the firing of the internal oninput event. Applies to the filtering as well since it occurs upon user input.
```
dimodi, line 89: When you say "Applies to the filtering AS WELL", what else does it apply to?

```suggestion
| `DebounceDelay` | `int` <br/> (`150`) | The milliseconds between the last typed symbol and the firing of the internal oninput event. Applies to the filtering as well, since it occurs upon user input.
```

components/numerictextbox/overview.md
```diff
@@ -78,7 +78,8 @@ The Blazor Numeric TextBox allows you to define your desired custom format throu
 
 | Attribute | Type and Default Value | Description |
 |-----------|------|-------------|
-|`Arrows`|`bool`<br/> Defaults to `true`|Whether the up/down spinner arrows (buttons) will be shown.|
+|`Arrows`|`bool`<br/> Defaults to `true`|Whether the up/down spinner arrows (buttons) will be shown.
```
dimodi, line 81:
```suggestion
| `Arrows` | `bool`<br /> (`true`) | Whether to show the up/down spinner arrows (buttons). |
```

dimodi (changes requested):
Consider the suggestions and if you like them, apply everywhere.

components/autocomplete/overview.md
ntacheva: Will reword it to cover scenarios when filtering is not enabled as well.

dimodi (approved):

PR #867, opened 2022-04-01
[KB] Add Grid Columns from Code

svdimitr:


dimodi:
Another real-world scenario will be adding columns without a user action. We should clarify what is the appropriate Razor component life cycle method for this (e.g. OnInitialized, OnParametersSet, OnAfterRender, etc.). Or anywhere after OnStateInit, etc.

dimodi (approved):

svdimitr:
> Another real-world scenario will be adding columns without user action. We should clarify what is the appropriate Razor component life cycle method for this (e.g. OnInitialized, OnParametersSet, OnAfterRender, etc.). Or anywhere after OnStateInit, etc.

According to me, adding Grid columns from code in the lifecycle hooks would be the same as defining them in the markup. This is a scenario where you want to add columns based on user action.

PR #869, opened 2022-04-01
docs(button,form): Document Button Form and Form Id parameters

dimodi:


components/form/overview.md
```diff
@@ -265,17 +263,18 @@ You can customize the editors further through the [form items]({%slug form-formi
 
 * `ColumnSpacing` - `string` - defines the space between the FormItems. See the [Columns]({%slug form-columns%}) article for more information.
 
+* `Id` - `string` - sets an `id` attribute to the `<form>` element. Use it in combination with the [`Form` parameter of a submit button]({%slug button-type%}). Set both parameters to the same `string` value. This allows submitting the form from a button, which is outside the form.
```
dimodi, line 266:
```suggestion
* `Id` - `string` - sets an `id` attribute to the `<form>` element. Use it together with the [`Form` parameter of a submit button]({%slug button-type%}). Set both parameters to the same `string` value. This allows submitting the form from a button, which is outside the form.
```

components/form/overview.md
```diff
@@ -265,17 +263,18 @@ You can customize the editors further through the [form items]({%slug form-formi
 
 * `ColumnSpacing` - `string` - defines the space between the FormItems. See the [Columns]({%slug form-columns%}) article for more information.
 
+* `Id` - `string` - sets an `id` attribute to the `<form>` element. Use it together with the [`Form` parameter of a submit button]({%slug button-type%}). Set both parameters to the same `string` value. This allows submitting the form from a button, which is outside the form.
```
dimodi, line 266:
```suggestion
* `Id` - `string` - sets an `id` attribute to the `<form>` element. It is possible to use it together with the [`Form` parameter of a submit button]({%slug button-type%}). Set both parameters to the same `string` value. This allows submitting the form from a button, which is outside the form.
```

xristianstefanov (approved):

PR #870, opened 2022-04-04
chore(splitbutton): Add specification updates and missing parameters

dimodi:


components/splitbutton/overview.md
```diff
@@ -90,6 +90,58 @@ The following table lists the SplitButton parameters, except those related to [b
 | `Title`| `string` | Sets the `title` attribute of the primary action element. |
 
 
+### SplitButton Popup Settings
```
ntacheva, line 93: May be we should omit the component name here, so the section name is consistent with the other popup settings sections - e.g. https://docs.telerik.com/blazor-ui/components/combobox/overview#popup-settings

```suggestion
### Popup Settings
```

ntacheva (approved):

PR #872, opened 2022-04-06
docs(form): Add KB for conditional validation

dimodi:


knowledge-base/form-conditional-validation.md
```diff
@@ -0,0 +1,50 @@
...
+
+* Use a [third-party validator that allows conditional validation]({%slug form-validation%}#fluent-validation).
+* Perform custom validation in the [Form's `OnSubmit` event]({%slug form-events%}#onsubmit).
+* Implement [remote (server-side) custom validation](https://github.com/telerik/blazor-ui/tree/master/form/remote-validation).
+* Use [`FormItem` Templates]({%slug form-formitems-template%}). Subscribe to the **change** handlers of the field editors to execute custom logic, show notifications, etc.
+* Implement a [conditional `DataAnnotation` attribute](https://stackoverflow.com/questions/26354853/conditionally-required-property-using-data-annotations). To see inline error messages next to the field editor, use the [`ValidationResult` overload that passes a **field name**](https://docs.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.validationresult.-ctor?view=net-6.0#system-componentmodel-dataannotations-validationresult-ctor(system-string-system-collections-generic-ienumerable((system-string)))). In this way the form validator will know which field has failed validation.
```
ntacheva, line 43: We have runnable examples in the linked resources for all other validation options but the last one with the conditional `DataAnnotation` attribute. May be it is worth adding one, it does not necessarily needs to be in this kb.

ntacheva (approved):

knowledge-base/form-conditional-validation.md
dimodi: It's general knowledge, so we can consider it if more people want to do this specific thing.

PR #944, opened 2022-05-12
Missing property in hardcoded data generation

uteschj:
The root directory needs the 'HasDirectories' property set to true for the component to display the child directories.


dimodi (approved):

dimodi:
Thanks, @uteschj !

PR #945, opened 2022-05-13
chore(grid):fixed kb example

xristianstefanov:


dimodi (changes requested):
Consider making the `.k-grid .k-link` combinator more specific.

dimodi (approved):

PR #954, opened 2022-05-18
chore(treeview):added example for a very common scenario

xristianstefanov:
- Added an example of how to expand a specific node programmatically. I have several tickets asking for such an example.

### DIFF
--- a/components/treeview/overview.md
+++ b/components/treeview/overview.md
@@ -91,7 +91,90 @@ The <a href="https://www.telerik.com/blazor-ui/treeview" target="_blank">Blazor 
     }
 }
 ````
+>caption Expand specific nodes programmatically
 
+````CSHTML
+@*To expand a specific node, the parent of this node needs to be expanded as well.*@
+
+<TelerikButton OnClick="ExpandSpecificNode">Expand Specific Node</TelerikButton>
+
+<TelerikTreeView Data="@FlatData"
+                 @bind-ExpandedItems="@ExpandedItems" />
+
+@code {
+    IEnumerable<TreeItem> FlatData { get; set; }
+    IEnumerable<object> ExpandedItems { get; set; } = new List<TreeItem>();
+
+    void ExpandSpecificNode()
+    {
+        FlatData = FlatData.Where(x => x.ParentId == null || x.Id == 4 || x.Id == 2).ToList();
+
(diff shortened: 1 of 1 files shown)


dimodi (changes requested):
This example is not real-world for two reasons -

* The data depends on which are the expanded items.
* The expanded items and their hierarchy is hard-coded.

The documentation needs a scenario, where the user can expand a random item (e.g. by its ID). If the item has parents that need to be expanded as well, the example should do that on the fly.

components/treeview/overview.md
```diff
@@ -91,7 +91,123 @@ The <a href="https://www.telerik.com/blazor-ui/treeview" target="_blank">Blazor
...
+                 @bind-ExpandedItems="@ExpandedItems"
+                 @bind-SelectedItems="@SelectedItems"
+                 SelectionMode="@TreeViewSelectionMode.Single" />
+
+@code {
+    public IEnumerable<TreeItem> FlatData { get; set; }
+    public IEnumerable<object> ExpandedItems { get; set; } = new List<TreeItem>();
+    public IEnumerable<object> SelectedItems { get; set; } = new List<TreeItem>();
+
+    int ItemToExpand { get; set; }
```
ntacheva, line 119:
```suggestion
    int? ItemToExpand { get; set; }
```

Use nullable int here, so there is no confusion if the user did not type anything in the NumericTextBox. Otherwise, by default the NumericTextBox will display 0, however there is no such item.

ntacheva:
I think that currently the example is hard to find. May be it will be better if we add it in a dedicated Expanded items article. We can  list some details for the property, link the events related to it and last but not least - how to programmatically control it.

xristianstefanov:
> I think that currently the example is hard to find. May be it will be better if we add it in a dedicated Expanded items article. We can list some details for the property, link the events related to it and last but not least - how to programmatically control it. Similar to https://docs.telerik.com/blazor-ui/components/panelbar/expanded-items - it looks cleaner.

We've previously discussed that this place seems to be the best as it is the first one the client sees when starts to explore the component. This way, I think, one section dedicated to only this would be unnecessary. There is not much information unseen in the other sections we can provide here if we expand the section list.

ntacheva (approved):

dimodi (approved):

PR #955, opened 2022-05-18
chore(grid):added an example for onrowrender with locked columns

xristianstefanov:
- Added an extra example for the OnRowRender event as I tracked several tickets asking for OnRowRender locked columns example.


ntacheva:
I couple issues I noticed:

- The snippet is also meant to demonstrate conditional formatting and now the condition is gone. I think it will be better to keep it.
- The hover color is not applied on my end.

ntacheva (changes requested):
Recommended changes listed above.

ntacheva (approved):

PR #1023, opened 2022-07-04
docs(panelbar):Revamp Overview

ntacheva:
Note: removed Reference section as there are no exposed methods for the component, so the section does not have value.

### DIFF
--- a/components/panelbar/overview.md
+++ b/components/panelbar/overview.md
@@ -154,7 +154,6 @@ The following table lists PanelBar parameters, which are not related to other fe
 ## Next Steps
 
 * [Binding PanelBar PanelBar to Data]({%slug panelbar-data-binding-overview%})
-
 * [Using the PanelBar to Navigate between Pages]({%slug panelbar-navigation%})
 
 
--- a/components/panelbar/overview.md
+++ b/components/panelbar/overview.md
@@ -12,18 +12,8 @@ position: 0
 
 The <a href="https://www.telerik.com/blazor-ui/panelbar" target="_blank">Blazor PanelBar component</a> displays [data]({%slug panelbar-data-binding-overview%}) ([flat]({%slug panelbar-data-binding-flat%}) or [hierarchical]({%slug panelbar-data-binding-hierarchical%})) in an accordion type structure. In addition to built-in [navigation capabilities]({%slug panelbar-navigation%}), you can navigate through the items and their children, define [templates]({%slug panelbar-templates%}), render text and [icons/images]({%slug panelbar-icons%}), and respond to [events]({%slug panelbar-events%}).
 
-
-
-#### In this article:
-
-* [Basics](#basics)
-
-* [Elements of a PanelBar Item](#elements-of-a-panelbar-item)
-
-
-## Basics
-
-To use a Telerik PanelBar for Blazor:
+ 
+## Creating Blazor PanelBar
 
 1. Add the `TelerikPanelBar` tag
 1. Provide a collection of models to its `Data` parameter (read more in the [Data Binding article]({%slug panelbar-data-binding-overview%}))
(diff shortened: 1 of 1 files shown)


components/panelbar/overview.md
```diff
@@ -121,22 +111,9 @@ To use a Telerik PanelBar for Blazor:
...
-<TelerikPanelBar @ref="@PanelBarReference">
-</TelerikPanelBar>
-
-@code {
-    private Telerik.Blazor.Components.TelerikPanelBar PanelBarReference { get; set; }
-}
-````
+To show any items, the Blazor PanelBar requires a data source that you can provide through the `Data` property. The PanelBar allows you to display the items both as flat data and hierarchically. [Read more about the Blazor PanelBar data binding...]({%slug panelbar-data-binding-overview%})
 
 ## Elements of a PanelBar Item
```
dimodi, line 118: This section can be misleading. When not using templates, "Header" and "Content" are two separate items. We should explain this in some other way. We can also rename the section to "Templates"

components/panelbar/overview.md
```diff
@@ -12,18 +12,8 @@ position: 0
...
-* [Basics](#basics)
-
-* [Elements of a PanelBar Item](#elements-of-a-panelbar-item)
-
-
-## Basics
-
-To use a Telerik PanelBar for Blazor:
+ 
+## Creating Blazor PanelBar
```
dimodi, line 16: Point 3 is not necessary in this case. We can only mention that such customization is possible, if the model class is using non-default property names. We can mention 2-3 default property names in point 2 or 3.

Probably we should describe the default property names first, and then add a step for binding the PanelBar to a collection of such items.

dimodi (changes requested):
The basic steps will bring more value if we mention default property names and the possibility to customize them.

The section about templates can be confusing in the part where we say one item has header and content.

components/panelbar/overview.md
```diff
@@ -121,32 +110,52 @@ To use a Telerik PanelBar for Blazor:
...
+
+The following table lists PanelBar parameters, which are not related to other features on this page. Check the [PanelBar API Reference](https://docs.telerik.com/blazor-ui/api/Telerik.Blazor.Components.TelerikPanelBar) for a full list of properties, methods and events.
+
+@[template](/_contentTemplates/common/parameters-table-styles.md#table-layout)
+
+| Attribute | Type and Default Value | Description |
+|----------|----------|----------|
+|`Class`| `string` |The CSS class that will be rendered on the main wrapping element of the component.|
+|`ExpandedItems`| `IEnumerable<Object>` | A collection of the expanded PanelBar items. Supports two-way binding. Read more in the [ExpandedItems]({%slug panelbar-expand-items%}) article.|
+|`ExpandMode`| `PanelBarExpandMode` enum <br/> (`PanelBarExpandMode.Multiple`) | Whether the PanelBar will allow single or multiple items to be expanded at a time. Read more in the [ExpandMode]({%slug panelbar-expand-items%}#expandmode) section. |
```
dimodi, line 151:
```suggestion
|---|---|---|
| `Class` | `string` | The CSS class that will be rendered on the main wrapping element of the component.|
| `ExpandedItems` | `IEnumerable<Object>` | A collection of the expanded PanelBar items. Supports two-way binding. Read more in the [ExpandedItems]({%slug panelbar-expand-items%}) article. |
| `ExpandMode` | `PanelBarExpandMode` enum <br/> (`PanelBarExpandMode.Multiple`) | Whether the PanelBar will allow single or multiple items to be expanded at a time. Read more in the [ExpandMode]({%slug panelbar-expand-items%}#expandmode) section. |
```

components/panelbar/overview.md
```diff
@@ -121,32 +110,52 @@ To use a Telerik PanelBar for Blazor:
...
+
+
+## Next Steps
+
+* [Binding PanelBar PanelBar to Data]({%slug panelbar-data-binding-overview%})
+
```
dimodi, line 157:
```suggestion
```

dimodi (approved):

PR #1048, opened 2022-07-18
docs(grid): Update @key values in dynamic columns KB

dimodi:
Based on https://www.telerik.com/forums/should-key-be-used-explicitly-with-telerik-blazor

### DIFF
--- a/knowledge-base/grid-dynamic-columns-with-static.md
+++ b/knowledge-base/grid-dynamic-columns-with-static.md
@@ -41,14 +41,16 @@ When I switch the number of dynamic columns or the column definitions, the comma
 <TelerikRadioGroup Data="@ColumnOptions"
                    Value="@CurrentColumns"
                    Layout="@RadioGroupLayout.Horizontal"
-                   OnChange="@( (object newValue) => ChangeGridColumns((int)newValue) )" />
+                   OnChange="@( (object newValue) => ChangeGridColumns((int)newValue) )">
+</TelerikRadioGroup>
 
-<TelerikGrid TItem="@GridItem" @ref="@GridRef"
-             OnRead="@OnGridRead">
+<TelerikGrid @ref="@GridRef"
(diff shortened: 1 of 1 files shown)


knowledge-base/grid-dynamic-columns-with-static.md
```diff
@@ -105,7 +111,7 @@ Typical scenarios that can exhibit such column position change are binding to `D
 
 To make a specific Grid column appear always last:
 
-* Set the [`@key`](https://docs.microsoft.com/en-us/aspnet/core/blazor/components/?view=aspnetcore-6.0#use-key-to-control-the-preservation-of-elements-and-components) directive of all columns to a unique value. This will help the Blazor framework distinguish the columns and their client-side rendering.
+* Set the [`@key`](https://docs.microsoft.com/en-us/aspnet/core/blazor/components/?view=aspnetcore-6.0#use-key-to-control-the-preservation-of-elements-and-components) directive of all columns to a unique value. This will help the Blazor framework distinguish the columns and their client-side rendering. Note that the `@key` of the command column should not remain the same when the number of dynamic columns changes.
```
svdimitr, line 114: Maybe we should use a more imperative language, like `**must** not remain the same....`. Currently, I think, It makes us uncertain if it should or should not remain the same.

svdimitr (approved):

knowledge-base/grid-dynamic-columns-with-static.md
dimodi, line 114:
```suggestion
* Set the [`@key`](https://docs.microsoft.com/en-us/aspnet/core/blazor/components/?view=aspnetcore-6.0#use-key-to-control-the-preservation-of-elements-and-components) directive of all columns to a unique value. This will help the Blazor framework distinguish the columns and their client-side rendering. Note that the `@key` of the command column must not remain the same when the number of dynamic columns changes.
```

PR #1054, opened 2022-07-19
docs(common): Add documentation about binding to cloud services

dimodi:


svdimitr (approved):

marin-bratanov:
I highly recommend adding a paragraph to the intro that says we are data source agnostic and what that means and that it is up to the app to deal with that data and we have nothing to do with it or support the actual connections or third party tooling.

PR #1075, opened 2022-07-27
Including [FromForm] in the Remove signature

bjswick33:
After following the controller example and some debugging I had to add [FromForm] to the Remove signature before the upload would hit the controller action.

CLA Form below has been signed/submitted.

### DIFF
--- a/components/upload/overview.md
+++ b/components/upload/overview.md
@@ -116,7 +116,7 @@ namespace MyBlazorApp.Controllers
         }
 
         [HttpPost]
-        public ActionResult Remove(string files) // must match RemoveField
+        public ActionResult Remove([FromForm] string files) // must match RemoveField
         {
             if (files != null)
             {


dimodi:
Hello @bjswick33 ,

Thanks for your contribution!

Normally, `[FromForm]` should not be required and I have Blazor Server and WebAssembly test projects that both work without it. Perhaps your scenario involves [model binding](https://docs.microsoft.com/en-us/aspnet/core/mvc/models/model-binding) or some other setup, which brings the need to add `[FromForm]`. This thread contains a few comments that may ring a bell -

[Asp.net core model doesn't bind from form](https://stackoverflow.com/questions/40172051/asp-net-core-model-doesnt-bind-from-form)](https://stackoverflow.com/questions/40172051/asp-net-core-model-doesnt-bind-from-form)

We can include `[FromForm]` in a note in our documentation, but we must be clear about what exactly triggers the need for this attribute? Thank you in advance for your cooperation!

bjswick33:
@dimodi 

Thanks for the input. I've run through my code and I'm really only using the sample code within my application. I'm running the latest version of everything and was wondering if the need for [FromForm] for the remove was needed since my upload was within an <EditForm>, but after testing outside of that the code still will not fire unless the attribute is added. Are your examples public on GH so I can bring them down? Are they using .NET 6 and latest versions?

In any case I think if the code example does not include the attribute some sort of message below in the documentation may need to be included for those having a similar action when trying to use the remove (attribute not required for save). We needed to do some searching when we found a discussion on the forms where someone did this and it's working.

dimodi:
Hi Brun,

Here is a runnable Blazor Server app that you can compare with your implementation. Uploading and removing works without `[FromForm]`.

[UploadTest.zip](https://github.com/telerik/blazor-docs/files/9253074/UploadTest.zip)

Let me know if you find what makes the difference.

bjswick33:
Thanks @dimodi. I'll have to take a look in the next day or so. The application will not build and has a number of errors. Main thing jumping out at me is we are .NET 6.0 and this app cannot target 6? We are also doing a blazor server (hosted) application. I'll get back to you in a day or so.

dimodi:
That's strange. The test app targets .Net 6 and I specifically made it from scratch, so that it is runnable as is.

bjswick33:
Yeah No problem. I only gave it a few minutes. Pretty busy today so can't 100% focus on it today. I'll let you know what I find.

bjswick33:
I'm sorry, but I have not. I've been continued working on the application and have not had time to set aside to test your example. I'm going to set aside some time tomorrow and will provide an update. Sorry it's taken this long.

dimodi:
Thanks, Brun.

In this case, I suggest that we close this for now. It will look strange to document a statement that sounds a bit vague about its applicability and relevance. If by chance another customer sheds light on this, we will revisit the topic.

PR #1081, opened 2022-07-28
docs(gantt): New Date properties in the Gantt TooltipTemplate

xristianstefanov:


dimodi (approved):
It's a bit peculiar to have documentation in the form of an example only. Perhaps we can consider changing this in the future.

xristianstefanov:
> It's a bit peculiar to have documentation in the form of an example only. Perhaps we can consider changing this in the future.

I agree we may think about it in the future.

PR #1097, opened 2022-08-05
docs(grid): Update programmatic export docs

ntacheva:


components/grid/export/csv.md
```diff
@@ -97,7 +97,7 @@ Optionally, you can also set the `GridCsvExport` tag settings under the `GridExp
 You can programmatically invoke the export feature of the Grid, by using the following methods exposed on the `@ref` of the Grid:
 
 * `SaveAsCsvFileAsync` - `ValueTask` - sends the exported CSV file to the browser for download
-* `ExportToCsvAsync` - `Task<MemoryStream>` - returns the exported data as a memory stream
+* `ExportToCsvAsync` - `Task<MemoryStream>` - returns the exported data as a `MemoryStream`. The stream itself is finalized, so that the resource does not leak. To read and work with the stream, clone its available binary data to a new `MemoryStream` instance.
```
dimodi, line 100: Maybe we should clarify that this workaround is only necessary if one wishes to work with stream methods, e.g. reading in chunks, etc. Otherwise one can just take the whole content and do something with it, right?

dimodi (approved):

components/grid/export/csv.md
ntacheva: What do you mean by "take the whole content and do something with it"? The stream is closed, so you cannot even read it. I cannot think of possible action with the stream that will not require its reading. I'm wondering why would you need the stream in this case?

PR #1108, opened 2022-08-15
Rendering changes

svdimitr:
The documentation for the rendering changes


upgrade/rendering-changes/list.md
```diff
@@ -0,0 +1,135 @@
...
+description: See which versions have rendering changes so you can take them into account when upgrading.
+page_title: List of Versions with Rendering Changes
+slug: versions-with-rendering-changes
+position: 0
+---
+
+# List of Telerik UI for Blazor Versions with Rendering Changes
+
+
+As the Telerik UI for Blazor matures we will continuously optimize the rendering and the usage of CSS classes. The rendering changes will affect the styling of the components if the business application overrode the built-in CSS rules. This article will track the version where such changes are introduced.
```
dimodi, line 12: Names are not used with articles, unless they play the role of an adjective:
* incorrect - "the Telerik UI for Blazor matures"
* correct - "the Telerik UI for Blazor components mature"

```suggestion
As Telerik UI for Blazor matures, we may optimize the HTML rendering and the usage of CSS classes. The rendering changes will affect the styling of the components if the application uses custom CSS that overrides the built-in styles.

This article will track the versions where such changes were introduced.
```

_contentTemplates/common/good-styling-practices.md
```diff
@@ -0,0 +1,3 @@
+#good-styling-practices
+The Telerik UI for Blazor components render as regular HTML elements. The best way to customize their appearance is by using CSS. To change the visual representation of the components easily you can use the wide variety of exposed properties and events that add custom CSS classes to different segments of the components. Following the good styling practices will also protect the business application from the affects of the [rendering changes]({%slug versions-with-rendering-changes%}).
```
dimodi, line 2: Link https://docs.telerik.com/blazor-ui/knowledge-base/common-theme-customization-options

upgrade/rendering-changes/list.md
```diff
@@ -0,0 +1,135 @@
...
+* [3.4.0]({%slug changes-in-3-4-0%})
+
+## Good Styling Practices
+
+@[template](/_contentTemplates/common/good-styling-practices.md#good-styling-practices)
+
+>caption Provide some custom CSS rules to the header and cells of the Grid. In the `Advisable` tab you can see the good practices in action. 
+
+<div class="skip-repl"></div>
+````Advisable
```
dimodi, line 23: How about "recommended" and "not recommended" ? "Inadvisable" sounds a bit exotic.

If you need to use single words, then perhaps "prefer" + "avoid"

upgrade/rendering-changes/3-4-0.md
```diff
@@ -0,0 +1,37 @@
...
+
+# Rendering Changes in 3.4.0
+
+## "k-state" CSS classes
+
+Previous to the 3.4.0 version we used `k-state-*` CSS classes to denote that the state of an HTML element happened. As part of the 3.4.0 release, we remove the `state` segment of that classes. 
+
+>caption A list of all changed "k-state" classes
+
+| Pre-3.4.0        | 3.4.0      |
```
dimodi, line 17: Didn't we change some of these for 3.0?

upgrade/rendering-changes/3-4-0.md
```diff
@@ -0,0 +1,37 @@
...
+| k-state-hovered  | k-hovered  |
+| k-state-selected | k-selected |
+| k-state-disabled | k-disabled |
+| k-state-invalid  | k-invalid  |
+| k-state-active   | k-active   |
+
+
+## Dialog
+
+* Added `k-dialog-wrapper` CSS class to the topmost HTML element.
```
dimodi, line 31: It will be good to mention which is the element (e.g. `div.k-dialog` if that's the one), so that a person can more easily see what new selectors will be possible to use.

dimodi (changes requested):
* Link the CSS Customization KB article.
* Clarify in more simple and actionable terms what are the good styling practices.

upgrade/rendering-changes/3-4-0.md
svdimitr: I have seen an issue for 3.4.0 where I took the list of classes. Maybe I must investigate where exactly these classes were changed.

upgrade/rendering-changes/3-4-0.md
svdimitr: It is just a div.k-dialog-wrapper, no other classes, the topmost div

styling-and-themes/good-styling-practices.md
```diff
@@ -0,0 +1,98 @@
...
+
+# Good Styling Practices
+
+@[template](/_contentTemplates/common/good-styling-practices.md#good-styling-practices)
+
+## Examples
+
+This section show to use the good styling practices to customize the appearance of the Telerik UI for Blazor components.
+
+````Grid
```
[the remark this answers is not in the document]
svdimitr: So, the general thinking here is that our clients cant customize something by using a custom CSS class (through a parameter/event we exposed) we did not mean that part of the component to be customized. If such a case pops up and we think it's valid we can think of a way to provide more accessible customization options. Overriding our CSS classes is not sustainable in the long run.

upgrade/rendering-changes/3-4-0.md
dimodi:
```suggestion
* Added `k-dialog-wrapper` CSS class to the outermost HTML element, which holds the modal overlay and the dialog instance.
```

styling-and-themes/good-styling-practices.md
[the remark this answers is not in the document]
dimodi: That's true, but I think that a lot of customer scenarios involve customizing nested elements, which are not directly accessible via Class parameters. Here is a simple example from today - "how do I remove the padding of the Grid toolbar"? If our documentation conveys the idea "always use your own CSS classes", customers may start asking "How do I set a custom class to the Grid toolbar". Instead, we can recommend to use custom classes whenever possible.

_contentTemplates/common/good-styling-practices.md
dimodi:
```suggestion
The Telerik UI for Blazor components render as regular HTML elements. The [best way to customize their appearance is by using CSS]({%slug common-kb-theme-customization-options%}). When using CSS overrides, we recommend to set custom CSS classes via the exposed component parameters and events whenever possible. Following such styling practices will also protect the business application from [rendering changes]({%slug versions-with-rendering-changes%}) in future versions.
```

dimodi (approved):
Some pending suggestions are not resolved (committed or discarded yet). It may also be good idea to link the Good Styling Practices from a few popular KBs about custom styling.

PR #1134, opened 2022-09-01
[3.7] docs(selects):Adaptive rendering docs

ntacheva:


common-features/adaptive-rendering.md
```diff
@@ -0,0 +1,67 @@
...
+````CSHTML
+// NOTE: The configurations below includeс only the DropDownList, but is applicable to all of teh above listed components
+
+// Adapts to the screen size to use the appropriate rendering.
+<TelerikDropDownList ... AdaptiveMode="@AdaptiveMode.Auto" ... />
+````
+
+# Rendering specifics
+
+When you set the `AdaptiveMode` to `Auto`, the component will take the screen size into consideration to use the appropriate rendering. The different rendering essentially targets the popup element of the component and how it will be displayed to the user.
```
svdimitr, line 49: The different rendering <del>essentially</del> targets the popup element of the component and how it will be displayed to the user.

svdimitr (approved):

PR #1137, opened 2022-09-02
[3.6.0] OnItemRender event documentation

svdimitr:


components/autocomplete/events.md
```diff
@@ -227,6 +228,62 @@ When using `OnRead`, make sure to set `TItem` and `TValue`.
...
+The `OnItemRender` event fires when each item in the collection passed to the `Data` parameter of the AutoComplete renders. 
+
+The event handler receives as an argument an `AutoCompleteItemRenderEventArgs<TItem>` object that contains:
+
+@[template](/_contentTemplates/common/parameters-table-styles.md#table-layout)
+
+| Property | Description                              |
+|----------|------------------------------------------|
+| `Item`   | The current item that renders in the AutoComplete. |
+| `Class`  | The custom CSS class that will be added to the item.     |
```
dimodi, line 242: We should either use ASCII art properly (which requires maintenance when the table contents changes), or remove it completely. This table was somewhere in the middle.

```suggestion
| Property | Description |
| --- | --- |
| `Item` | The current item that renders in the AutoComplete. |
| `Class` | The custom CSS class that will be added to the item. |
```

components/autocomplete/events.md
```diff
@@ -227,6 +228,62 @@ When using `OnRead`, make sure to set `TItem` and `TValue`.
...
+    }
+</style>
+
+<TelerikAutoComplete Data="@Suggestions"
+                     OnItemRender="@OnItemRenderHandler"
+                     ValueField="@( nameof(SuggestionsModel.Suggestion) )"
+                     @bind-Value="@TheValue" />
+
+@code {
+    string TheValue { get; set; }
```
dimodi, line 260: `TheValue` is not according to our style guide. How do you distinguish properties if you have two components on the page?

```suggestion
    private string TheValue { get; set; }
```

components/autocomplete/events.md
```diff
@@ -227,6 +228,62 @@ When using `OnRead`, make sure to set `TItem` and `TValue`.
...
+<style>
+    .disabled-item{
+        pointer-events: none;
+        color: gray;
+    }
+</style>
+
+<TelerikAutoComplete Data="@Suggestions"
+                     OnItemRender="@OnItemRenderHandler"
+                     ValueField="@( nameof(SuggestionsModel.Suggestion) )"
```
dimodi, line 256:
```suggestion
                     ValueField="@nameof(SuggestionsModel.Suggestion)"
```

components/autocomplete/events.md
```diff
@@ -227,6 +228,62 @@ When using `OnRead`, make sure to set `TItem` and `TValue`.
...
+                     @bind-Value="@TheValue" />
+
+@code {
+    string TheValue { get; set; }
+
+    public void OnItemRenderHandler(AutoCompleteItemRenderEventArgs<SuggestionsModel> args)
```
dimodi, line 262:
```suggestion
    private void OnItemRenderHandler(AutoCompleteItemRenderEventArgs<SuggestionsModel> args)
```

components/autocomplete/events.md
```diff
@@ -227,6 +228,62 @@ When using `OnRead`, make sure to set `TItem` and `TValue`.
...
+
+@[template](/_contentTemplates/common/parameters-table-styles.md#table-layout)
+
+| Property | Description                              |
+|----------|------------------------------------------|
+| `Item`   | The current item that renders in the AutoComplete. |
+| `Class`  | The custom CSS class that will be added to the item.     |
+
+````CSHTML
+@* Disable an item in the AutoComplete *@
```
dimodi, line 245: Consider custom styling based on a model value, rather than "disabled" item. It's not truly disabled, because you can select it via keyboard navigation. This will generate support or negative docs votes.

svdimitr: We are implementing this to respond to the Disabled Items support in Selects (https://feedback.telerik.com/blazor/1474264-add-support-for-disabled-items).

svdimitr: To add to this, maybe we can enhance the CSS rules and in the public feedback item showcase OnitemRender + cancel the OnClose event so it truly disables an item.

components/autocomplete/events.md
```diff
@@ -227,6 +228,62 @@ When using `OnRead`, make sure to set `TItem` and `TValue`.
 }
 ````
 
+## OnItemRender
+
+The `OnItemRender` event fires when each item in the collection passed to the `Data` parameter of the AutoComplete renders.
```
dimodi, line 233: Consider shorter and simpler sentences. This one contains a very long subordinate clause that separates two related words - "item" and "renders". It's also more technical that necessary. For example:

* The `OnItemRender` event fires when each item in the `Data` collection renders.
* The `OnItemRender` event fires when each item in the AutoComplete dropdown renders.

It's also worth clarifying if the event fires every time the dropdown is opened or not. What about filtering scenarios, when the visible items change all the time?

dimodi (changes requested):
All suggestions apply for all components.

components/autocomplete/events.md
dimodi: You can change the selected item with the keyboard even without opening the dropdown. So `OnItemRender` is irrelevant here and the requested feature still does not exist.

dimodi: See ticket 1576049 that shows "more" "disabled" items. This can be a KB, while the `OnItemRender` example in the docs can focus on styling customizations to keep it simple, plus optional link to the KB.

components/autocomplete/events.md
svdimitr: It will fire again if the items in the popup are re-rendered (e.g. Filtering is used). So it truly fires each time an item renders.

On the topic of the sentence, I have noticed that we are using longer sentences when explaining things in the docs. I totally agree with you that this can be shortened by removing the subordinate clause.

components/autocomplete/events.md
```diff
@@ -227,6 +228,63 @@ When using `OnRead`, make sure to set `TItem` and `TValue`.
 }
 ````
 
+## OnItemRender
+
+The `OnItemRender` event fires when each item in the AutoComplete dropdown renders.
+
+The event handler receives as an argument an `AutoCompleteItemRenderEventArgs<TItem>` object that contains:
```
dimodi, line 235:
```suggestion
The event handler argument is an `AutoCompleteItemRenderEventArgs<TItem>` object that contains:
```

PR #1152, opened 2022-09-13
Accessibility Documentation

jivanova:


dimodi:
In relation to the heavy usage of tables here, it may be a suitable time to remove the [notorious fixed table layout from docs-seed](https://github.com/telerik/docs-seed/blob/master/_assets/css/_styles-layout.scss#L1122).

svdimitr (approved):

PR #1170, opened 2022-10-11
[3.7] docs(tabstrip): Refresh method

xristianstefanov:


components/tabstrip/overview.md
```diff
@@ -83,6 +83,51 @@ The following parameters enable you to customize the appearance of the Blazor Ta
...
+<TelerikTabStrip @ref="@TabRef">
+    <CityPopulation TabRef="TabRef" />
+</TelerikTabStrip>
+
+@code {
+    public TelerikTabStrip TabRef { get; set; }
+}
+````
+````CityPopulation.razor
+@* This code snippet showcases an example usage of the Refresh() method. *@
```
dimodi, line 108:
```suggestion
@* This code snippet demonstrates usage of the TabStrip Refresh() method. *@
```

components/tabstrip/overview.md
```diff
@@ -83,6 +83,51 @@ The following parameters enable you to customize the appearance of the Blazor Ta
 | `Width`   | `string` | The width of the component. You can set the Width parameter to any of the [supported units]({%slug common-features/dimensions%}). |
 | `Height`  | `string` | The height of the Component. You can set the `Height` parameter to any of the [supported units]({%slug common-features/dimensions%}). |
 
+## TabStrip Reference and Methods
+
+The `TabStrip` methods are accessible through its reference.
+
+| Method | Description |
+| --- | --- |
+| `Refresh` | Redraws the component. |
```
dimodi, line 92: We can mention what is the benefit of this method. For example, the alternative is to call `StateHasChanged()` on the parent, which will achieve the same result, but will re-render more content, and not just the TabStrip.

dimodi (approved):

PR #1185, opened 2022-10-25
chore(listview): fix refresh article

xristianstefanov:


dimodi (approved):
All accessors should be `private`, except for Blazor life cycle methods, that must be `protected`.

PR #1205, opened 2022-11-18
docs(selects): Revamp Filter articles

dimodi:


components/autocomplete/filter.md
```diff
@@ -10,85 +10,80 @@ position: 10
...
-@* On the first keystroke, there will be no suggestions, then you will only get "Developer" and "Designer" as you write "de" *@
-
-<TelerikAutoComplete Data="@Suggestions" @bind-Value="@TheValue"
-                     Filterable="true" MinLength="2"
-                     Placeholder="Write 'de' to see the filtering" ClearButton="true" />
+<ul>
+    <li>
+        <label>
+            Choose filter operator:
+            <select @bind="FilterOperator">
```
ntacheva, line 38: We may use a DropDownList here for consistency and a better look.

ntacheva (changes requested):
The MultiColumnComboBox Filter article is not refactored - we should consider it as well as it falls under the selects category.

ntacheva (approved):

PR #1232, opened 2022-12-02
[3.8] TreeView OnItemRender

svdimitr:


dimodi (changes requested):
* Pair the `__Items` and `___ItemsChanged` parameters on adjacent lines in the TreeView declaration.
* If you like string logging in the page UI, then replace all `Console.WriteLine` statements.
* Move the `<style>` tag above the `@code` section.
* What is `ChosenItem` used for?

dimodi (approved):

PR #1267, opened 2023-01-04
[KB] Focus a FormItem programmatically

svdimitr:


knowledge-base/form-focus-formitem.md
```diff
@@ -0,0 +1,96 @@
+---
+title: Set Programmatic Focus to a FormItem
```
dimodi, line 2: The customer has used the keyword "initial", which is not present anywhere in this article.

knowledge-base/form-focus-formitem.md
```diff
@@ -0,0 +1,96 @@
...
+        <tr>
+            <td>Product</td>
+            <td>Telerik UI for Blazor</td>
+        </tr>
+    </tbody>
+</table>
+
+## Description
+
+How to set programmatic focus to a `<FormItem>`?
```
dimodi, line 26: Since we don't focus the `FormItem` itself, but an input component inside it, we can enhance the article description, so that it contains more keywords. For example, add another line, which says "How to focus a specific input component in the form by default" or any variation of that.

knowledge-base/form-focus-formitem.md
```diff
@@ -0,0 +1,96 @@
...
+
+## Solution
+
+The `<FormItem>` is an abstraction of the real editor that is rendered in the browser. To set programmatic focus you must:
+
+1. Use the [FormItem Template]({%slug form-formitems-template%}) to provide the desired editor.
```
dimodi, line 32:
```suggestion
1. Use a [FormItem Template]({%slug form-formitems-template%}) to provide the desired editor.
```

knowledge-base/form-focus-formitem.md
```diff
@@ -0,0 +1,96 @@
...
+        [Required]
+        [Display(Name = "Date of Birth")]
+        public DateTime? DOB { get; set; }
+    }
+}
+````
+
+## Notes
+
+* The **earliest reliable programmatic focus** can occur in `OnAfterRenderAsync` and with some delay. The reason is that OnAfterRenderAsync is fired when the DOM tree is built, but **before** the HTML output is actually rendered in the browser. After the event is fired, the .NET runtime sends the HTML to the browser. The `FocusAsync` method relies on `JSInterop`, which in turn relies on the component to be rendered in the browser.
```
dimodi, line 96:
```suggestion
* The **earliest reliable programmatic focus** can occur in `OnAfterRenderAsync` and with some delay. The reason is that `OnAfterRenderAsync` fires when the DOM tree is built, but **before** the HTML output is actually rendered in the browser. After the event is fired, the .NET runtime sends the HTML to the browser. The `FocusAsync` method relies on `JSInterop`, which in turn relies on the component to be rendered in the browser.
```

dimodi (approved):
Some SEO improvements may be nice.

PR #1281, opened 2023-01-16
docs(dateinputs): Document R1 2023 typing parameters

dimodi:


components/dateinput/keyboard-typing.md
```diff
@@ -0,0 +1,165 @@
...
+````
+
+See the [complete example](#example) below.
+
+
+## AutoSwitchParts
+
+`AutoSwitchParts` sets if the caret will **move automatically to the next date format segment** when the user completes the current one. It's `true` by default.
+
+* When `true`, the the caret will **not** move for days and months if the segment value is equal to `1` or `2`, because the value may not be complete yet. In this case, the user should press an arrow key or an "auto switch key" to advance.
```
ntacheva, line 79:
```suggestion
* When `true`, the caret will **not** move for days and months if the segment value is equal to `1` or `2`, because the value may not be complete yet. In this case, the user should press an arrow key or an "auto switch key" to advance.
```

components/dateinput/overview.md
```diff
@@ -10,119 +10,144 @@ position: 0
...
+
 
 ## Creating Blazor DateInput
 
-1. Add the `TelerikDateInput` tag to your razor page.
-1. Bind a `DateTime` object to the component
-1. Optionally, provide custom `Format`, `Min` and `Max` values
+1. Add a `TelerikDateInput` tag to your razor page.
+1. Bind the `Value` parameter to a [`DateTime` or `DateTime?` object](#nullable-datetime). The parameter supports two-way binding.
+1. (optional) Set the [`Format`, `Min` and `Max` parameters](#date-input-parameters).
```
ntacheva, line 20:
```suggestion
1. (optional) Set the [`Format`, `Min` and `Max` parameters](#dateinput-parameters).
```

components/dateinput/overview.md
```diff
@@ -10,119 +10,144 @@ position: 0
...
->caption Basic DateInput with custom format, min and max
+>caption Basic Telerik Blazor DateInput
 
 ````CSHTML
-@dateInputValue
-<br />
-
-<TelerikDateInput @bind-Value="@dateInputValue" Format="dd MMMM yyyy"
-                  Min="@Min" Max="@Max">
+<TelerikDateInput @bind-Value="@DateValue"
```
ntacheva, line 26: The sample does not include a demonstration of the 4th bullet. Maybe we can enable some of the new parameters.

components/dateinput/overview.md
```diff
@@ -10,119 +10,144 @@ position: 0
...
+The component provides multiple parameters, which control the [caret placement, two-digit year values and the auto-correct behavior of the Date Input]({%slug dateinput-keyboard-typing%}).
 
-You can find more options for customizing the Date Input styling in the [Appearance article]({%slug dateinput-appearance%}).
-
-@[template](/_contentTemplates/date-inputs/format-placeholders.md#format-placeholder)
-
-## DateTime and Nullable DateTime
-
-The behavior of the component will depend on the type of field it is bound to, and this can result in different user experience and values that it will output:
+### Styling and Appearance
```
ntacheva, line 103: I suggest moving the `Class` and `Width` parameters to this section similar to https://docs.telerik.com/blazor-ui/components/datepicker/overview#styling-and-appearance

ntacheva (approved):

components/dateinput/keyboard-typing.md
```diff
@@ -0,0 +1,165 @@
...
+* When `false`, the user should always advance to the next segment manually with the arrow keys or any of the `AutoSwitchKeys`.
+
+See the [example](#example) below.
+
+
+## TwoDigitYearMax
+
```
ntacheva, line 86: Maybe we should add a note that this setting applies to the input only and the users can select a year that exceeds the `TwoDigitYearMax` from the calendar(applies for pickers). If they want to restrict the selection as well, they need to specify `Max`.

components/dateinput/overview.md
dimodi: This makes sense, but I have a few concerns about this -
* The Parameters subsections will become too many
* Class and Width are just two parameters that do not justify their own section.
* There are other parameters related to appearance, which are not mentioned in the Overview, so everything becomes a bit messy to me.

components/dateinput/keyboard-typing.md
dimodi: I think this can deceive developers and create confusion in users. Indeed, you can select any year via the Calendar, but if you then decide to change it via the keyboard, the century will change and the user will not know about this.

So I think we should advertise `TwoDigitYearMax` only in the restricted use case. Nevertheless, I enhanced the documentation, based on your comment.

PR #1299, opened 2023-01-20
docs(treeview): Add KB for disabled TreeView

dimodi:


svdimitr (approved):

svdimitr:
Maybe we can extend this example, by adding the OnItemRender event and making the nodes visually disable by graying them out a bit (like the k-disabled class).

PR #1307, opened 2023-02-03
docs(tilelayout): add tile visible parameter description

yanisslav:


xristianstefanov (approved):

svdimitr:
Maybe we can change the description a bit by not using the root of the word. In this case, visible is pretty self-explanatory, however we should keep that in mind for future documentation.

PR #1345, opened 2023-02-27
chore(scheduler): fixed slot sample based on feedback

xristianstefanov:


components/scheduler/templates/slot.md
```diff
@@ -46,6 +46,8 @@ The `context` of the template is a `SchedulerSlotTemplateContext` object that co
 ````CSHMTL
 @* Use the AllDaySlotTemplate and SlotTemplate *@
 
+@using Telerik.FontIcons
```
dimodi, line 49: Is this necessary? Customers should have it in `_Imports`. If REPL requires it, notify Kris to import the namespace by default.

dimodi (approved):

components/scheduler/templates/slot.md
xristianstefanov: Ops, you are right.

PR #1357, opened 2023-03-13
[4.1] docs(map): add article for custom map markers

xristianstefanov:
Related to: https://github.com/telerik/blazor/issues/5908.

### DIFF
--- a/components/map/layers/marker.md
+++ b/components/map/layers/marker.md
@@ -12,6 +12,15 @@ position: 5
 
 The marker functionality allows you to add points in the map. These points are defined by geographical position in the map and can show useful information to the user in a tooltip.
 
+Sections in this article:
+
+* [Creating Map Marker Layer](#creating-map-marker-layer)
+* [Marker Template](#marker-template)
+* [Marker Shapes](#marker-shapes)
+* [Marker Tooltip Settings](#marker-tooltip-settings)
+
+## Creating Map Marker Layer
+
 **To configure a Map Layer of type Marker:**
 
 1. Add the `TelerikMap` tag.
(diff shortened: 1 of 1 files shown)


components/map/layers/marker.md
```diff
@@ -81,20 +90,105 @@ The following example demonstrates how to configure the Map Marker Layer.
 }
 ````
 
->caption The result from the above code snippet.
+## Marker Template
+
+To customize the marker appearance, set the `Template` parameter in the corresponding marker inner tag - `MapLayerMarkerSettings`. It accepts `string` and `RenderFragment`.
```
Stamo-Gochev, line 95:
```suggestion
To customize the marker appearance, set the `Template` parameter in the corresponding marker inner tag - `MapLayerMarkerSettings`.
```
It doesn't work with a `RenderFragment` - see https://github.com/telerik/blazor/issues/5381#issuecomment-1438537859

> In addition, we cannot easily plug a RenderFragment (even if the Map is not written in C#) as we do with the tooltip, because the tooltip element is a single one - and we display it in an async way (on a particular user interaction), while the rendering of the markers is async - and it is preferable to avoid the N number of calls between JS and C# just to render the markers - and there can be a lot of markers. Considering other features that we can add in the future, e.g. marker clustering, such things will be even harder to be handled.

components/map/layers/marker.md
```diff
@@ -81,20 +90,105 @@ The following example demonstrates how to configure the Map Marker Layer.
...
+                  UrlTemplate="@UrlTemplate">
+        </MapLayer>
+
+        <MapLayer Type="@MapLayersType.Marker"
+                  Data="@MarkerData2"
+                  LocationField="@nameof(MarkerModel.LatLng)"
+                  TitleField="@nameof(MarkerModel.Title)">
+                  <MapLayerMarkerSettings Template="#= dataItem.LatLng #">
+                      <MapLayerMarkerSettingsTooltip>
+                          <Template>
```
Stamo-Gochev, line 127: We can split the examples into 2 sections - one for marker template and one for marker tooltip template.

Stamo-Gochev (changes requested):

components/map/layers/marker.md
xristianstefanov: I don't seem to have understood the comment correctly. It will be fixed.

components/map/layers/marker.md
xristianstefanov: I will just remove the marker tooltip template, as there is already a sample on that page that shows it.

Stamo-Gochev (approved):

dmihaylo (approved):
As a general recommendation, consider removing the "Sections in this article:" internal table of contents (TOC) even for longer topics. 
The approach is not consistent (not in all articles), duplicates unnecessary the right-hand TOC, and is not really visible as it's part of the paragraph and looks like yet another list of options.

PR #1364, opened 2023-03-21
Add configuration wizard article

yanisslav:
Related to: [telerik/blazor - 6277](https://github.com/telerik/blazor/issues/6277)


dmihaylo (approved):
As a general recommendation for the whole documentation:

Move all troubleshooting scenarios as separate articles to the KB section. 
In the way the are exposed now, it's impossible for users to identify the problem they have through organic search as we have part of them inside a conceptual topic, and the other scenarios grouped in a separate topic too.

If you want to still make them visible after the removal, you can refer to them with links similar to the [Known Issues in AJAX docs](https://docs.telerik.com/devtools/aspnet-ajax/getting-started/known-issues).

PR #1387, opened 2023-04-07
Fix broken code snippets in the Drawer Templates article

svdimitr:


dimodi (approved):
We shouldn't need @using Telerik.FontIcons, because the Getting Started and Icons documentation requires it.

PR #1440, opened 2023-05-09
kb(grid):filter by selected items

ntacheva:

PR #1442, opened 2023-05-09
Review grid filter by selection kb

yordan-mitev:
Review suggestions for kb(grid):filter by selected items #1440


ikoevska (commented):
Ignored.

PR #1521, opened 2023-06-30
kb: update change grid icons kb

yanisslav:


dimodi (changes requested):
The SVG customization doesn't work in Safari. We can:

* mention this explicitly (also check other browsers)
* suggest using font icons instead

Also, consider consistent comments in the Razor file (`@**@` vs `<!-- -->`).

Link the Icons documentation page.

yordan-mitev (approved):

dimodi (approved):

PR #1537, opened 2023-07-12
docs(inputs): add readonly parameter in the parameters section

yanisslav:
Related to: telerik/blazor#6965


components/textbox/overview.md
```diff
@@ -54,6 +54,7 @@ The Blazor TextBox provides various parameters to configure the component:
 | `Class` | `string` | The custom CSS class to be rendered on the `<span class="k-textbox">` element. |
 | `DebounceDelay` | `int` | Specifies the time in milliseconds between the last typed symbol and the updating of the value. The default value is 150ms. |
 | `Enabled` | `bool` <br /> `true` | Whether the `input` is enabled. |
+| `ReadOnly` | `bool` | If set to `true`, the component will be readonly and will not allow user input. The component is not readonly by default and allows user input. |
```
dimodi, line 57: We don't have to specify the default value, if it matches the default value of the type.

```suggestion
| `ReadOnly` | `bool` | The `readonly` attribute of the `<input />`. If `true`, the component will not allow user input. |
```

dimodi (approved):
If you like my suggestion, apply it everywhere.

yordan-mitev (approved):

PR #1561, opened 2023-07-28
docs(localization): Remove obsolete info and refactor code

dimodi:
Based on ticket 1617119

### DIFF
--- a/globalization/localization.md
+++ b/globalization/localization.md
@@ -35,10 +35,11 @@ Telerik provides and supports the default English strings. The other language st
 
 ## How to Enable Localization in Your App
 
-When localizing a Blazor app, make sure you are familiar with the way localization works in the framework. You can start from the following resources:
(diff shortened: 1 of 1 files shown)


globalization/localization.md
```diff
@@ -65,120 +66,34 @@ The necessary steps are to:
...
+````CS
 var builder = WebApplication.CreateBuilder(args);
 
-// Add services to the container.
-builder.Services.AddRazorPages();
-builder.Services.AddServerSideBlazor();
+// ...
+
+// AddTelerikBlazor() registers the built-in service that includes only the default English labels.
+// The actual localizer for the Telerik components should be registered after the line below.
```
yordan-mitev, line 83: should -> must

yordan-mitev (approved):

globalization/localization.md
dimodi, line 83:
```suggestion
// The actual localizer for the Telerik components must be registered after the line below.
```

PR #1606, opened 2023-08-31
Update KB: Filter and Edit Enum in Grid

ntacheva:
Add `[Display(Name = "...")]` based on a customer feedback.

### DIFF
--- a/knowledge-base/grid-filter-edit-enum.md
+++ b/knowledge-base/grid-filter-edit-enum.md
@@ -31,7 +31,7 @@ For versions prior to `2.15.0`, you could use a [custom editor template]({%slug 
 
 As of **2.15.0**, the grid provides enum filtering and editing through dropdowns out-of-the-box without any additional code.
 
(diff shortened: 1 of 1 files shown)


knowledge-base/grid-filter-edit-enum.md
```diff
@@ -31,12 +31,17 @@ For versions prior to `2.15.0`, you could use a [custom editor template]({%slug
 
 As of **2.15.0**, the grid provides enum filtering and editing through dropdowns out-of-the-box without any additional code.
 
+To control how the enum values will be shown, decorate them with [`Display` attributes](https://learn.microsoft.com/en-us/dotnet/api/microsoft.openapi.attributes.displayattribute) and set the `Name` property. The Grid will use the `Name` value for the editors. If you do not use a `[Display(Name = "...")]` attribute, the Grid will take the string representation of the enum value.
```
dimodi, line 34:
```suggestion
To control how each enum value will show in the filter DropDownList, decorate it with the [`Display` attribute](https://learn.microsoft.com/en-us/dotnet/api/microsoft.openapi.attributes.displayattribute) and set the `Name` property. The Grid will use the `Name` value also for the default column editor. If you do not use a `[Display(Name = "...")]` attribute, the Grid will take the string representation of the enum value.
```

dimodi (approved):

yordan-mitev (approved):

PR #1650, opened 2023-09-18
docs(grid): Improve row dragging documentation and example

dimodi:


components/grid/row-drag-drop.md
```diff
@@ -132,48 +132,57 @@ This section contains the following examples:
 
 ### Drag and Drop a Row between Grids
 
-When you drap and drop items from one Grid to another, the `OnRowDrop` event fires for both Grid instances, so you can update their data sources. All Grid instances must be bound to the same model. It is also possible to [bind the Grids to different models, but they must be derived from the same interface]({%slug grid-kb-drag-drop-different-models%}).
+When you drap and drop items from one Grid to another, the `OnRowDrop` event fires for both Grid instances, so you can update their data sources. All Grid instances must be bound to the same model type. It is also possible to [bind the Grids to different models, but they must be derived from the same interface]({%slug grid-kb-drag-drop-different-models%}).
```
yordan-mitev, line 135:
```suggestion
When you drag and drop items from one Grid to another, the `OnRowDrop` event fires for both Grid instances, so you can update their data sources. All Grid instances must be bound to the same model type. It is also possible to [bind the Grids to different models, but they must be derived from the same interface]({%slug grid-kb-drag-drop-different-models%}).
```

components/grid/row-drag-drop.md
```diff
@@ -132,48 +132,57 @@ This section contains the following examples:
...
 
-````CSHTML
-@* Drag a row from one Grid and Drop it in the other *@ 
+The target drop area in the Grid is the `<table>` element. Users cannot drop in the empty space below the last table row and this includes the `NoDataTemplate` too. There are two ways to prevent possible confusion and enhance the UX:
```
yordan-mitev, line 137:
```suggestion
The target drop area in the Grid is the `<table>` element. Users cannot drop items in the empty space below the last table row and this includes the `NoDataTemplate` too. There are two ways to prevent possible confusion and enhance the UX:
```

yordan-mitev (approved):

PR #1653, opened 2023-09-19
docs(common): Add information about new .NET version features

dimodi:


system-requirements.md
```diff
@@ -59,6 +59,10 @@ A **compatible .NET version** is one that is *no longer* (or *not yet*) official
 
 Telerik UI for Blazor is committed to the currently supported official versions of .NET. We evaluate the compatibility with preview .NET versions on a regular basis and provide support for new .NET versions after their official release becomes available.
 
+### Features From Newer .NET Versions
+
+Telerik UI for Blazor can include features from newer .NET versions in its source code only after it drops support for older .NET versions. For example, .NET 7 will be supported by mid 2024 and .NET 6 will be supported by late 2024. As a result, the Telerik Blazor components can include .NET 8 features in 2025 at the earliest.
```
yordan-mitev, line 64: - There seems to be a slight inconsistency with the tense. So we will drop support for .NET7 before we drop support for .NET6?
- Use "until" instead of "by" as it removes the ambiguity.
- "Features From Newer .NET Versions" -> ".NET Versions Lifecycle"

dimodi: * Yes, [.NET 6 is an LTS version, which will be supported after .NET 7](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core)
* Done.
* The proposed heading sounds too general and implies different meaning, which is closer to the previous sections in the article. The new content in this PR aims to answer questions like "When you will introduce feature X from .NET version Y?", which we start getting around each new .NET release.

yordan-mitev (approved):

PR #1733, opened 2023-11-06
docs(grid): fix grid knowledge base example

YanushevGeorgi:
One of the custom CSS selectors did not work because it had lower specificity than one of our CSS selectors.


yordan-mitev (approved):

dimodi (changes requested):
Increase the specificity in another way.

PR #1837, opened 2024-01-05
docs(filemanager): fix docs samples

xristianstefanov:
This enhancement is a result of this discussion: https://github.com/telerik/blazor/issues/7111#issuecomment-1877087480

### DIFF
--- a/components/filemanager/data-binding/flat-data.md
+++ b/components/filemanager/data-binding/flat-data.md
@@ -42,7 +42,8 @@ You must also provide the correct value for the `HasDirectories` field - for the
 @code {
     private List<FlatFileEntry> FileManagerData = new List<FlatFileEntry>();
 
-    private string DirectoryPath { get; set; } = string.Empty;
+    private string RootPath { get; set; } = "root-folder-path";
+    private string DirectoryPath { get; set; } = "directory-folder-path";
 
     // fetch the FileManager data
     protected override async Task OnInitializedAsync()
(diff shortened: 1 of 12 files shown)


components/filemanager/overview.md
```diff
@@ -195,133 +194,133 @@ The <a href = "https://www.telerik.com/blazor-ui/file-manager" target="_blank">B
...
+            gridDesign
+        };
 
         return await Task.FromResult(files);
     }
 }
 ````
 
->The `Path` of each `FileManagerEntry` should end with the `Name` of the data item, mirroring the format used in actual file systems.
+>Each `FileManagerEntry`'s `Path` needs to finish with the data item's `Name`, following the style used in real file systems. Also, there should be a constant root path as the topmost path segment to show the Breadcrumb right. If there's no root path given, we consider the first folder is the root. However, this causes an issue because the Breadcrumb doesn't recognize it and shows the `Home` icon instead.
```
yanisslav, line 323: Perhaps, we should avoid adding specific details about the discussed problem, as there might be other issues beyond just the breadcrumb. 

Additionally, the root path segments is important not just because the breadcrumb. Clients can have more than one 'root' directory and can change it by adjusting the 'Path' parameter value of the FileManager. 

For instance, you can bind the FileManager to a collection of files with different topmost 'root' directories (for example 'C:/...' and 'D:/...'). Thus you can programmatically change the `Path` of the FileManager from 'C:/...' to 'D:/...' to display other files.

===

In my opinion, adding a brief sentence emphasizing that the topmost file's path should include a root path at the start is enough.

xristianstefanov: Done

components/filemanager/data-binding/flat-data.md
```diff
@@ -42,7 +42,8 @@ You must also provide the correct value for the `HasDirectories` field - for the
 @code {
     private List<FlatFileEntry> FileManagerData = new List<FlatFileEntry>();
 
-    private string DirectoryPath { get; set; } = string.Empty;
+    private string RootPath { get; set; } = "root-folder-path";
+    private string DirectoryPath { get; set; } = "directory-folder-path";
```
yanisslav, line 46: Is there a reason for the `DirectoryPath` to be different from the `RootPath`? As I see, there are no files in this directory, which means the `FileManager` will be initially empty. See: https://blazorrepl.telerik.com/mSkPEXOU23Yy45qA57

It is probably a good idea to set its initial value to match the `RootPath`. Thus initially the FileManager will show the root directory. 

This suggestion is applicable to all examples.

```suggestion
    private string RootPath { get; set; } = "root-folder-path";
    private string DirectoryPath { get; set; } = "root-folder-path"; // The initially opened directory will be the root directory.
```

components/filemanager/overview.md
```diff
@@ -195,133 +194,133 @@ The <a href = "https://www.telerik.com/blazor-ui/file-manager" target="_blank">B
...
+            gridDesign
+        };
 
         return await Task.FromResult(files);
     }
 }
 ````
 
->The `Path` of each `FileManagerEntry` should end with the `Name` of the data item, mirroring the format used in actual file systems.
+>Each `FileManagerEntry`'s `Path` needs to finish with the data item's `Name`, following the style used in real file systems. Also, there should be a predefined root path as the topmost path segment.
```
yanisslav, line 323:
```suggestion
>Each `FileManagerEntry`'s `Path` needs to finish with the data item's `Name`, following the style used in real file systems. Additionally, the paths of the topmost directories and files should start with a predefined root path.
```

xristianstefanov: I made it even more simple: "Also, ensure the top-level directories and files start with a predefined root path."

yanisslav (approved):

PR #1840, opened 2024-01-08
chore(window): list specifics for using a custom icon for a built-in …

ntacheva:


components/window/actions.md
```diff
@@ -65,6 +65,7 @@ Action buttons expose the following properties:
 }
 ````
 
+>Setting a custom icon for a built-in action is not supported - the Window will override it and use the default SVG icon for the corresponding built-in action. If you need to specify a custom icon, use a [custom action](#custom-actions) instead of the built-in one.
```
radkostanev, line 68:
```suggestion
>Setting a custom icon for a built-in action is not supported - the Window will override it and use the default icon for the corresponding built-in action. If you need to specify a custom icon, use a [custom action](#custom-actions) instead of the built-in one. 
```

Although SVG is the default icon type, the Window will still render built-in font icons as long as this is how the app is configured, hence it might be best to omit the icon type.

components/window/actions.md
```diff
@@ -75,9 +76,12 @@ You can create a custom action icon and you must provide its `OnClick` handler.
...
 Custom actions can call C# directly
 
-<TelerikWindow Visible="true">
+<link href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/6.5.1/css/all.min.css" rel="stylesheet" />
+
+<TelerikWindow Visible="@WindowVisible">
 	<WindowActions>
-		<WindowAction Name="MyAction" Icon="@SvgIcon.Gear" OnClick="@MyCustomActionHandler" />
+		<WindowAction Name="MyAction" Title="MyAction" Icon="@SvgIcon.Gear" OnClick="@MyCustomActionHandler" />
+		<WindowAction Name="CustomClose" Title="CustomClose" Icon="@("fas fa-x")" OnClick="@(()=> WindowVisible = false)" />
```
radkostanev, line 84: With the current implementation, the `Name` parameter of the `Window` is practically redundant when it comes to custom actions. In other words, it is currently used as a weakly typed enumeration, as the available values are `Close`, `Maximize`, and `Minimize`. 

Do you see value in mentioning this, and perhaps even omitting the `Name` declaration from custom actions?

components/window/actions.md
yordan-mitev, line 68: is not supported _as_ the Window will override it...

components/window/actions.md
ntacheva: Agreed, I'll omit the `Name` declaration.

radkostanev (approved):

PR #1882, opened 2024-01-24
docs(selects): document LoaderShowDelay parameter

ntacheva:
https://github.com/telerik/blazor/issues/8369#issuecomment-1908797748


ikoevska (approved):
🚀 

As a general recommendation - it might make sense to explore content reuse for repetitive text (in this case, the description). This would let you create/edit in one place and reuse wherever you need it. Read more about how to do it [here](https://jekyllrb.com/docs/includes/).

Contact me, if you'd like to explore how to do it here.

PR #1885, opened 2024-01-25
docs(grid): Document expand persistence of hierarchy editing

dimodi:


components/grid/editing/overview.md
```diff
@@ -449,6 +449,7 @@ There are a few considerations to keep in mind with the CUD operations of the gr
 
 @[template](/_contentTemplates/common/grid-treelist-editing-notes.md#grid-treelist-data-operations-while-editing)
 
+* When editing a master row in a [hierarchy Grid]({%slug components/grid/features/hierarchy%}), the respective `DetailTemplate` will collapse unless you [override the `Equals()` method of the master data item class]({%slug grid-kb-editing-in-hierarchy%}).
```
yordan-mitev, line 452: The `## Notes` section of the article has become so long that you could split it into two or even three new articles. My concern here is that if the page visitors don't know exactly what they are looking for, they will never find it.

knowledge-base/grid-edit-in-hierarchy.md
```diff
@@ -10,65 +10,84 @@ res_type: kb
...
 ## Description
 
-This article showcases how to **Create**, **Update** and **Delete** items in both the main and nested Grids in a [hierarchy]({%slug components/grid/features/hierarchy%}).
+This article shows how to create, update, and delete items in the main (master) and nested (detail) Grids in a [hierarchy]({%slug components/grid/features/hierarchy%}).
 
 
 ## Solution
 
-The [CRUD operations]({%slug components/grid/editing/overview%}) are independent from the Hierarchy. Each Grid has it`s own handlers.
+The Grid CRUD operations are independent from the Hierarchy. Each Grid performs editing separately and has its own handlers.
```
yordan-mitev, line 30:
```suggestion
The Grid CRUD operations are independent of the hierarchy. Each Grid performs editing separately and has its own handlers.
```

knowledge-base/grid-edit-in-hierarchy.md
```diff
@@ -10,65 +10,84 @@ res_type: kb
...
+1. [Configure editing for the main Grid]({%slug components/grid/editing/overview%}).
+1. Define a [`DetailTemplate`]({%slug components/grid/features/hierarchy%}) and configure the nested Grid inside it, including editing.
+1. To persist the expanded state of a `DetailTemplate` while editing its parent item, override the `Equals()` method of the master data item class. This is supported from version **5.1.0**.
 
-You can set the `EditMode` of the nesting grid to either `Popup`, `Inline`, `Incell`. By default it is `Inline`.
+The example below shows the separate CUD event handlers and also how to get the parent data item in the nested Grid handlers in case you need this information. The `UpdateOrder`, `CreateOrder` and `DeleteOrder` handlers receive the `productItem` in order to access the data item from the parent Grid. If you don't need access to the main Grid data, there is no need to pass the context through a lambda function.
```
yordan-mitev, line 36:
```suggestion
The example below shows the separate CUD event handlers and also how to get the parent data item in the nested Grid handlers. The `UpdateOrder`, `CreateOrder`, and `DeleteOrder` handlers receive the `productItem` in order to access the data item from the parent Grid. If you don't need access to the main Grid's data, don't pass the context through a lambda function.
```

yordan-mitev (approved):

components/grid/editing/overview.md
dimodi: I agree, but we'll leave that for a major revamp that is necessary for other reasons too.

PR #1944, opened 2024-02-21
kb(grid): fix search in numeric and date fields kb sample

xristianstefanov:
The change is needed so the arrow keys correctly navigate the input caret inside the toolbar.


dimodi (approved):

ikoevska (commented):
Reviewed.

PR #1963, opened 2024-03-06
chore(localization): refactor article structure

ntacheva:


yordan-mitev (commented):
I suggest splitting the article into two by taking out the tutorial from here and moving it to another article.

dimodi (approved):
I approve these changes, as they improve the current content and structure.

However, we should think about an additional revamp, which should resolve a few existing issues:

* We walk through the process for Blazor Server only. Perhaps we can move to .NET 8 Web App instead (regardless of the render mode).
* We should find a way to not discuss general knowledge and focus only on our bits.
* The blazor-ui apps must be updated.

ntacheva:
I am merging this now and we will track the rest of the proposed changes in this issue https://github.com/telerik/blazor/issues/9039.

PR #1981, opened 2024-03-19
Wizard step template

svdimitr:


svdimitr:
The main purpose of this article is to showcase that there is a template that allows our clients to take control over the rendering of the Steps, part of the Stepper in the Wizard component. 

We do not have a real-world example, more than we have for the Step Template in the Stepper component. 

The clients will not expect to look for the information in Stepper.

radkostanev (approved):

PR #2008, opened 2024-03-29
Rating documentation

xristianstefanov:
Related to: https://github.com/telerik/blazor/issues/8125

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -415,6 +415,8 @@ navigation:
         title: "QRCode"
      "*/radiobutton":
         title: "RadioButton"
+     "*rating":
+        title: "Rating"
      "*gauges/radial":
         title: "Radial"
      "*rangeslider":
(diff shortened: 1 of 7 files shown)


ikoevska (approved):
Approving so you can merge after addressing the minor issues noted in the review.

However, although you have documented these parameters, it might be worth discussing (and showcasing with a code snippets) the use case for a "a read-only rating" (with a label next to it). 

Some other things that come to mind and I don't really find here - is data binding a thing for ratings (that is, can you display a rating from a data base, and update it there when the user changes the selection)? Can you animate the selection process and how? When would you use the throttle setting? Can you replace the numeric values with words (say, on a scale from "didn't like it at all" to "it was superb" instead of 1 to 5 - much like Goodreads are doing - I'm guessing you would use templates for that?).

PR #2022, opened 2024-04-05
Update nuget.md

LanceMcCarthy:

### DIFF
--- a/installation/nuget.md
+++ b/installation/nuget.md
@@ -124,7 +124,6 @@ To edit a `NuGet.Config` file and add the Telerik feed, you need to:
     </configuration>
     ````
 
-
 ## Use NuGet API Key
 
 There are two ways to authenticate with the Telerik NuGet server when you add the Telerik NuGet source [with the .NET CLI](#use-the-net-cli) or [edit the `NuGet.Config` file manually](#edit-the-nugetconfig-file):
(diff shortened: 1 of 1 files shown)


installation/nuget.md
```diff
@@ -158,6 +157,7 @@ The `Telerik.UI.for.Blazor` NuGet package and most of its dependencies reside on
 </packageSourceMapping>
 ````
 
+> **Important!** Make sure you update the names `nuget.org` and `TelerikOnlineFeed` to match what you are using in the packageSources section. If these do not match, you will get a "package not found" error (because the incorrectly named source does not exist).
```
dimodi, line 160:
```suggestion
> Make sure you update the names `nuget.org` and `TelerikOnlineFeed` to match what you are using in the `packageSources` section. If these do not match, you will get a "Package not found" error, because the incorrectly named source does not exist.
```

LanceMcCarthy: I didnt realize this was an either-or commit resolution. Unresolving and choosing the 2nd one. My only concern is we rely on the reader to stop and understand what we are referring to by "keys", but we can gauge that over time.

LanceMcCarthy: committed the 2nd option manually

yordan-mitev (approved):

PR #2048, opened 2024-04-17
docs(AIPrompt): Remove events from parameters table

yordan-mitev:
Removing the auto-layout as the table doesn't render correctly on small screens.
image


dimodi (approved):

ikoevska (commented):
Reviewed.

PR #2067, opened 2024-04-23
docs(common): Fix incorrect mentioning of primitive types

dimodi:
Triggered by docs feedback. Refer to:

https://learn.microsoft.com/en-us/dotnet/api/system.type.isprimitive?view=net-8.0#remarks

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/built-in-types

So, our mentioning of "primitive" types is more restrictive than it should be.

### DIFF
--- a/_contentTemplates/common/observable-data.md
+++ b/_contentTemplates/common/observable-data.md
@@ -19,9 +19,9 @@ The Observable collections fire the `CollectionChanged` event only when their `A
 #refresh-data
 In Blazor, the framework will fire the `OnParametersSet` event of a child component (which is how child components can react to outside changes) only when it can detect a change in the object it receives through the corresponding parameter (like `Data` for the data sources of Telerik components). This detection works as follows:
 
-  * For primitive types (such as numbers, strings), this happens when their value changes.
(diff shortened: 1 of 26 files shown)


yordan-mitev (commented):
When referring to external resources, such as Microsoft, we recommend using HTML to open the link in a new tab. This helps reduce the bounce rate. Consider replacing the default markdown linking that you've used with `<a href="https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/value-types" target="_blank">value types</a>`.

dimodi:
> When referring to external resources, such as Microsoft, we recommend using HTML to open the link in a new tab. This helps reduce the bounce rate. Consider replacing the default markdown linking that you've used with `<a href="https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/value-types" target="_blank">value types</a>`.

The Blazor documentation doesn't follow this practice, so my suggestion is to switch all ~ 2500 markdown external links at once when we decide to do so.

yordan-mitev (approved):

PR #2093, opened 2024-05-10
docs(Chip): add aria-label and id parameters

Tsvetomir-Hr:
Related to : [#8861](https://github.com/telerik/blazor/issues/8861)

### DIFF
--- a/components/chip/overview.md
+++ b/components/chip/overview.md
@@ -59,7 +59,7 @@ The table below lists the Chip parameters. Also check the [Chip API Reference](/
 
 | Parameter | Type and Default&nbsp;Value | Description |
 | --- | --- | --- |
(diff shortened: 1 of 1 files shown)


components/chip/overview.md
```diff
@@ -59,9 +59,11 @@ The table below lists the Chip parameters. Also check the [Chip API Reference](/
 
 | Parameter | Type and Default&nbsp;Value | Description |
 | --- | --- | --- |
+| `AriaLabel` | `string` | Maps to the `aria-label` attribute. Use  this parameter if the text that labels the component is not visible. |
```
dimodi, line 62: attribute of which HTML element?

```suggestion
| `AriaLabel` | `string` | Maps to the `aria-label` attribute. Use this parameter if the text that labels the component is not visible. |
```

dimodi (commented):
There is no need for the `merge-to-production` label. We don't have RC for the docs site yet.

ikoevska (approved):

PR #2095, opened 2024-05-10
kb(Form): keep form items order

xristianstefanov:
From 1647539


knowledge-base/form-keep-fields-order.md
```diff
@@ -0,0 +1,105 @@
+---
+title: Keep Form Items Rendering Order after Show/Hide
+description: How to keep the form item rendering order upon toggling its visibility.
```
ikoevska, line 3:
```suggestion
description: How to keep the form item rendering order when changing the form item visibility.
```

knowledge-base/form-keep-fields-order.md
```diff
@@ -0,0 +1,105 @@
...
+## Description
+
+This knowledge base article answers the following questions:
+
+* How do I keep the rendering order of Form items?
+* Is it possible to toggle the visibility of a Form item and keep its rendering order instead of showing it as last?
+
+## Solution
+
+The `<FormItems>` tag doesn't render Form items directly. This tag is used only to register `<FormItem>` instances. The Form finds out about the conditional field item last, so it puts it last.
```
ikoevska, line 32: I feel that some context might be lacking here. What is the "conditional field item" mentioned at the end (I'm guessing this is the form item that gets shown/hidden and moves around the form)? You probably need to provide some extra explanation, such as "form items whose visibility is subject to change are considered conditional field items" (or whatever makes sense).

knowledge-base/form-keep-fields-order.md
```diff
@@ -0,0 +1,105 @@
...
+
+Thus, the solution is to use the `TelerikFormItemRenderer` as it defines the rendering place of each form item.
```
ikoevska, line 34: Would that be "instead of" or "along with"? 
```suggestion
To preserve the rendering place of each form item, you need to use the `TelerikFormItemRenderer` instead of the `<FormItems>` tag.
```

knowledge-base/form-keep-fields-order.md
```diff
@@ -0,0 +1,105 @@
...
+    {
+        public string TextField { get; set; }
+        public int ValueField { get; set; }
+    }
+}
+````
+
+## See Also
+
+* [Form Template for All Items]({%slug form-formitems-formitemstemplate%})
```
ikoevska, line 105: You might want to link to the other Form KBs that are available and to the Forms demo.

knowledge-base/form-keep-fields-order.md
xristianstefanov: Not technically correct so I adjusted it a little.

ikoevska (approved):

PR #2109, opened 2024-05-15
[Grid] Revamp the Virtual Scrolling article

svdimitr:
Related to: https://github.com/telerik/blazor/issues/9209

### DIFF
--- a/components/grid/virtual-scrolling.md
+++ b/components/grid/virtual-scrolling.md
@@ -18,8 +18,6 @@ You can also Virtually Scroll the Grid Columns. See the [Column Virtualization](
 
 ## Using Virtual Scrolling
 
-Setting a Value for the Height Parameter
-
 1. Set the `ScrollMode` parameter to `GridScrollMode.Virtual` (the default value is `Scrollable`).
 1. [Set the Height parameter](#setting-a-value-for-the-height-parameter).
 1. [Set the RowHeight parameter](#setting-a-value-for-the-rowheight-parameter).
--- a/components/grid/virtual-scrolling.md
+++ b/components/grid/virtual-scrolling.md
@@ -21,7 +21,7 @@ You can also use virtual scrolling for the Grid columns. See the [Column Virtual
 1. Set the `ScrollMode` parameter to `GridScrollMode.Virtual` (the default value is `Scrollable`).
 1. [Set the `Height` parameter](#setting-a-value-for-the-height-parameter).
 1. [Set the `RowHeight` parameter](#setting-a-value-for-the-rowheight-parameter).
-1. [Set the PageSize parameter](#setting-a-value-for-the-pagesize-parameter).
+1. [Set the `PageSize` parameter](#setting-a-value-for-the-pagesize-parameter).
 
 ## Setting a Value for the Height Parameter
 
--- a/components/grid/virtual-scrolling.md
+++ b/components/grid/virtual-scrolling.md
@@ -10,23 +10,39 @@ position: 60
 
 # Virtual Scrolling
 
-Virtual scrolling is an alternative to paging. Instead of using a pager, the user scrolls vertically through all records in the data source.
+Virtual scrolling provides an alternative to paging. Instead of utilizing a pager, the user scrolls vertically through all records in the data source.
 
-The same set of elements is reused to improve the rendering performance. While the next data is loading, a loading indicator is shown on the cells. If the user scrolls back up after scrolling down to a next set of rows, the previous data will be loaded anew from the data source, like with regular paging, but the scroll distance determines the data to be loaded.
+To enhance rendering performance, the Grid reuses the same set of HTML elements. As the next data loads, a loading indicator appears on the cells. If the user scrolls back up after scrolling down to the next set of rows, the previous data reloads from the data source, similar to regular paging, with the scroll distance determining the data to be loaded.
 
-You can also Virtually Scroll the Grid Columns. More information can be found in the [Column Virtualization]({%slug grid-columns-virtual%}) article.
+You can also Virtually Scroll the Grid Columns. See the [Column Virtualization]({%slug grid-columns-virtual%}) article for more information.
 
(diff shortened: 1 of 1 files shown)


svdimitr:
https://github.com/telerik/blazor-docs/pull/2109/commits/9134b115dd5ff41e2eeba9ee6c96cfe189f6eb6a#diff-6dfff5244aefd03df14e4bde914ab8ec7ec2d0480a510d8ceaa0c74fbe6089bcL27 - this is leftover from the previous article, I will delete it tomorrow.

components/grid/virtual-scrolling.md
```diff
@@ -79,54 +87,250 @@ To enable virtual scrolling:
...
+}
+````
+
+## Setting a Value for the PageSize Parameter
+
+Set the `PageSize` parameter to an `int` value. The `PageSize` determines how many rows are rendered at any given time and how many items are requested from the data source when loading data on demand. For optimal performance, use a page size that fills the grid's data viewport without being excessively large.
+
+## Considerations
+
+Virtualization primarily enhances client-side rendering performance and improves the user experience. However, it comes with the trade-off that certain features of the grid are not compatible with it. An alternative approach is to utilize [regular paging]({%slug components/grid/features/paging%}) combined with [manual data source operations]({%slug components/grid/manual-operations%}) to achieve the desired data retrieval performance.
```
yordan-mitev, line 321:
```suggestion
Virtualization primarily enhances client-side rendering performance and improves the user experience. However, it comes with the trade-off that certain features of the Grid are incompatible with it. An alternative approach is to utilize [regular paging]({%slug components/grid/features/paging%}) combined with [manual data source operations]({%slug components/grid/manual-operations%}) to achieve the desired data retrieval performance.
```

components/grid/virtual-scrolling.md
```diff
@@ -79,54 +87,250 @@ To enable virtual scrolling:
...
+
+List of the known limitations of the virtual scrolling feature:
```
yordan-mitev, line 323: Use a complete sentence instead of a fragment.
```suggestion
These are the known limitations of the virtual scrolling feature:
```

components/grid/virtual-scrolling.md
```diff
@@ -79,54 +87,250 @@ To enable virtual scrolling:
...
+
+* [Hierarchy]({%slug components/grid/features/hierarchy%}) is not supported.
+
+* [Grouping]({%slug components/grid/features/grouping%}) is not supported. [Loading Group Data On Demand]({%slug grid-group-lod%}) is supported, however.
 
 
 ## See Also
 
   * [Live Demo: Grid Virtual Scrolling](https://demos.telerik.com/blazor-ui/grid/virtual-scrolling)
+  * [Documentation: Selection in Grid with virtualized rows]({%slug components/grid/selection/overview%}#selection-in-grid-with-virtualized-rows)
```
yordan-mitev, line 333: * Use title case capitalization as in the first list item.
* As we are in the docs, you don't need to call this out.
```suggestion
  * [Selection in Grid with Virtualized Rows]({%slug components/grid/selection/overview%}#selection-in-grid-with-virtualized-rows)
```

components/grid/virtual-scrolling.md
```diff
@@ -79,54 +87,250 @@ To enable virtual scrolling:
...
+  * [Knowledge-Based Article: Virtual Scroll does not show all items]({%slug grid-kb-virtualization-many-records%})
```
yordan-mitev, line 334: Edit the "Knowledge Base Article" part in the next two anchor links as well.
```suggestion
  * [Knowledge Base Article: Virtual Scroll Does Not Show All Items]({%slug grid-kb-virtualization-many-records%})
```

dimodi (changes requested):
* The "Setting a value for the" part makes the headings too long and hard to scan when the reader looks in the right column. Consider headings that are just the property names or "Setting Height"
* The same consideration applies for "Knowledge base article: "
* There is too much attention and examples on setting the `Height`. Are there any specifics that customers must keep in mind, which are related to the feature? Do customers have problem with that? The only specific is the relationship between the `Height` and the `PageSize`. In addition, the "relative" example still applies a hard-coded height to the Grid, instead of `GridHeight`.
* There is a need for more clarifications about the `PageSize` - too few items will make skeletons show at all times. Larger `PageSize` will make skeletons appear less frequently, but too large `PageSize` will make the data requests inefficient.
* "The value of the RowHeight must be greater than **or equal to** the height of the **row** that the browser would normally render." Use only "row", because the row height is easier to see in the DOM inspector, and has no padding. You can also mention the exact minimum height number for the default theme just to give an idea, as customers tend to set too small values.
* "Grouping is not supported" is too general and negative statement. You can rephrase like "Grouping is supported only in load-on-demand mode" or "Grouping is supported only if LoadGroupsOnDemand=true" or something like that.
* To make sure the revamp serves the customers, find tickets in which we have linked the Grid virtualization article and see what issues customers had. This can give ideas for further improvements.

svdimitr:
@dimodi on each point:

* "Setting Height" - I agree it is better - I will change all the headings
* I spoke to Dancho about the length of "Knowledge base article:", the only reason I have left it like that is to be consistent as we tend to write "Documentation:" or "Live Demo:". I can remove the "Knowledge base article:" and leave only the KB title.
* The `Height` is one of the most asked for. People seem to like setting it to `Percentage` and other relative units. I have purposefully made as many examples. With [this commit](https://github.com/telerik/blazor-docs/pull/2109/commits/78253608e4fa7239653b9d21994f6217b0b83a09) I already fixed the value of the Height parameter in the relative example. 
* I have left out this information about the `PageSize` on purpose because there is a KB on the topic, that I have linked.
* I am not sure that I understand the suggestion. 
* I will rename the Grouping not supported bullet, your suggestion is better.
* I have gone through 20 tickets on the topic and the revamp covers these scenarios

yordan-mitev (approved):

PR #2124, opened 2024-05-20
docs(chart):add KB to link to export example

NansiYancheva:
Related to: https://github.com/telerik/blazor/issues/7634

### DIFF
--- a/knowledge-base/chart-export-to-image.md
+++ b/knowledge-base/chart-export-to-image.md
@@ -1,6 +1,6 @@
 ---
 title: Export Chart as Image
-description: How to export the Chart to image
+description: How to export the Chart as an image? How to export the rendered Chart as a JPG file? How to export Chart as an image in Blazor?
 type: how-to
 page_title: Export Chart as Image
 slug: chart-kb-export-to-image
@@ -23,7 +23,7 @@ res_type: kb
 
 ## Description
 
-How to export the Chart to an image? How to export the Chart to a JPG file?
+How to export the Chart as an image? How to export the Chart as a JPG file?
 
 
 ## Solution
--- a/knowledge-base/chart-export-to-image.md
+++ b/knowledge-base/chart-export-to-image.md
@@ -28,4 +28,4 @@ How to export the Chart as an image? How to export the Chart as a JPG file?
 
 ## Solution
 
-An example is available in the [JPG Export in the Browser with JS project](https://github.com/telerik/blazor-ui/tree/master/common/pdf-jpg-export-js). The project includes a `readme` file that contains guideline and information on the needed libraries for the export.
+An example is available in the [project for Chart JPG Export with JS](https://github.com/telerik/blazor-ui/tree/master/common/pdf-jpg-export-js). The project includes a `readme` file that contains guideline and information on the needed libraries for the export.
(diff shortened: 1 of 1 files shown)


yordan-mitev (commented):
The value of this KB is questionable at best as it lacks content. I would suggest reworking it and adding at least the primary steps required to export the Chart as an image. You could of course keep the link to the GitHub project with the complete implementation.

NansiYancheva:
> The value of this KB is questionable at best as it lacks content. I would suggest reworking it and adding at least the primary steps required to export the Chart as an image. You could of course keep the link to the GitHub project with the complete implementation.

@yordan-mitev 

The reasons for the KB appearing as it does are:

1. The main reason for creating the KB is the lack of a link to the GitHub project in the documentation. 
2. The GitHub project has instructions. We aim to avoid duplication of the guidelines (and their maintenance). The KB indicates that the GitHub project contains a README file with instructions.
3. We have another KB that looks the same https://docs.telerik.com/blazor-ui/knowledge-base/chart-print-chart-only
4. Additionally, we are planning to deprecate the GitHub repository and extract what is possible into KBs. In the meantime, we are trying to avoid duplicating guidelines while still utilizing available resources.

If these reasons are not acceptable, I will include the instructions from the GitHub project in the KB.

ntacheva:
We have this type of KB articles for most projects in the blazor-ui repo. Their main purpose is to provide a link to the corresponding project.

Indeed, we have previously discussed that such KBs do not have much value. In addition, a lot of the customers do not like the idea of having to browse to the repo and download the app in order to just test the specific scenario. That said, we will have an initiative for extracting all projects to KBs and deprecate the blazor-ui repo. This, however will take time and we better have the KB linking the app until then.

ntacheva (approved):

NansiYancheva:
@yordan-mitev please let us know if you are ok with the proposed approach

knowledge-base/chart-export-to-image.md
```diff
@@ -0,0 +1,31 @@
+---
+title: Export Chart to image
```
yordan-mitev, line 2:
```suggestion
title: Export Chart as Image
```

knowledge-base/chart-export-to-image.md
```diff
@@ -0,0 +1,31 @@
...
+description: How to export the Chart to image
```
yordan-mitev, line 3: The description must contain at least [100 characters](https://docs.telerik.com/style-guide/metadata).

knowledge-base/chart-export-to-image.md
```diff
@@ -0,0 +1,31 @@
...
+type: how-to
+page_title: Export Chart to image
```
yordan-mitev, line 5:
```suggestion
page_title: Export Chart as Image
```

yordan-mitev:
> @yordan-mitev please let us know if you are ok with the proposed approach

You are free to add the file as is. This, however, is not an actual KB and has no value.
- The main idea of a KB is to allow our clients to find solutions _on their own_. They cannot find a solution if it isn't described in an article that has sufficient content to allow search engines to index it.
- A KB that isn't indexed or ranked high serves no purpose if _you_ are the one who has to send the link to the client.

PR #2154, opened 2024-06-10
docs(Grid): fix typo in reoder columns

xristianstefanov:
Reported typo from client.

### DIFF
--- a/components/grid/columns/reorder.md
+++ b/components/grid/columns/reorder.md
@@ -14,7 +14,7 @@ The Grid lets the user reorder its columns by dragging their headers.
 
 To enable the column reordering, set the `Reorderable` parameter of the grid to `true`.
 
-To prevent the user from moving a certain column, set its own parameter `Reorderable="false"`. Note that the user can still re-arrange other columns around it.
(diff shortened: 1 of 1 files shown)


components/grid/columns/reorder.md
```diff
@@ -14,7 +14,7 @@ The Grid lets the user reorder its columns by dragging their headers.
 
 To enable the column reordering, set the `Reorderable` parameter of the grid to `true`.
 
-To prevent the user from moving a certain column, set its own parameter `Reordarable="false"`. Note that the user can still re-arrange other columns around it.
+To prevent the user from moving a certain column, set its own parameter `Reorderable="false"`. Note that the user can still re-arrange other columns around it.
```
yordan-mitev, line 17: * "its own" is redundant
* "Reorderable" modifies "parameter" and must precede it
```suggestion
To prevent the user from moving a certain column, set the column's `Reorderable` parameter to `"false"`. Note that the user can still re-arrange other columns around it.
```

dimodi (approved):

PR #2218, opened 2024-07-05
docs(textbox): Improve Appearance examples

dimodi:


yordan-mitev (approved):

components/textbox/appearance.md
```diff
@@ -29,28 +29,28 @@ You can increase or decrease the size of the TextBox by setting the `Size` attri
...
+>caption Built-in TextBox sizes
 
 ````CSHTML
 @{
     var fields = typeof(Telerik.Blazor.ThemeConstants.TextBox.Size)
-        .GetFields(System.Reflection.BindingFlags.Public | System.Reflection.BindingFlags.Static
-        | System.Reflection.BindingFlags.FlattenHierarchy)
+        .GetFields(System.Reflection.BindingFlags.Public
+            | System.Reflection.BindingFlags.Static
+            | System.Reflection.BindingFlags.FlattenHierarchy)
```
radkostanev, line 39: Have we considered omitting this reflection entirely in favor or hardcoded values? The pros of the reflection is that we can essentially use the same code for different examples and possibly allow us to showcase new values without modifying the examples. 

However, hardcoding the values offer transparency, as the user browsing the documentation can see all available values declared within the example.

Have we received any feedback in this regard?

radkostanev: I cannot come up with a good way to measure this, thus it might be best to just rely on a mental note and pay attention for any future feedback.

PR #2219, opened 2024-07-09
docs(Grid): update examples in kb articles

Tsvetomir-Hr:
Related to: https://github.com/telerik/blazor/issues/9612

### DIFF
--- a/knowledge-base/grid-filter-column-list.md
+++ b/knowledge-base/grid-filter-column-list.md
@@ -219,8 +219,7 @@ The example below includes two Grids - one for each `FilterMode`.
         MenuFilteredSpiceIds = new List<int>();
         await context.ClearFilterAsync();
 
-        // Since the filtering occurs outside the Grid,
-        // we need to clear the active filter style manually.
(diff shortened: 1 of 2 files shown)


knowledge-base/grid-filter-column-list.md
```diff
@@ -197,6 +218,10 @@ The example below includes two Grids - one for each `FilterMode`.
     {
         MenuFilteredSpiceIds = new List<int>();
         await context.ClearFilterAsync();
+
+        // Since the filtering occurs outside the Grid,
+        // we need to clear the active filter style manually.
```
ikoevska, line 223: Using we/our is not a good practice ([source](https://docs.telerik.com/style-guide/basic-rules-and-guidelines#pronouns)). Consider rewriting the entire comment into something like this:

`Because the filtering occurs outside of the Grid, you need to clear the active filter style manually.`
OR
`Because the filtering occurs outside of the Grid, the active filter style requires manual clearing.`

Tweak the recommendation as needed below as well.

ikoevska (approved):

xristianstefanov (approved):

PR #2337, opened 2024-08-12
docs(popover): Improve basic Popover example

dimodi:


components/popover/overview.md
```diff
@@ -15,28 +15,37 @@ The <a href = "https://www.telerik.com/blazor-ui/popover" target="_blank">Blazor
...
+1. Set the `AnchorSelector` parameter to a CSS selector, which points to the HTML element that the Popover will align with.
+1. Set the `ShowOn` parameter to an `PopoverShowOn` enum value to set which user interaction with the anchor will display the Popover automatically.
+1. [Obtain the component reference with the `@ref` attribute](#popover-reference-and-methods) to show and hide the Popover programmatically.
 1. Add the content to the `<PopoverContent>` child tag.
-1. Use the [`Show`](#popover-reference-and-methods) method to display the `<TelerikPopover>`.
-1. (optional) Add a title inside a `<PopoverHeader>` tag. HTML markup and child components are supported, too.
+1. (optional) Configure the Popover `Position` and `Offset`, or add a title inside the `<PopoverHeader>` tag.
 
->caption Basic configuration of the Telerik Popover for Blazor
+Steps 3 and 4 are optional, but you must implement at least one of them.
```
yordan-mitev, line 24: So they aren't optional. If you don't implement one of them, the component will not work correctly. I suggest combining them in a common step on the root level and then using a bullet list to present these steps. For example:
...
1. Show the Popover programmatically or automatically on user interaction by using at least one of these approaches:
   * step 3
   * step 4 
...

ntacheva (approved):

components/popover/overview.md
dimodi: Good suggestion, thanks.

PR #2450, opened 2024-10-18
kb(Notification): Add KB for single Notification instance

dimodi:
Related to https://github.com/telerik/blazor-ui/pull/308


ikoevska (approved):
Ideally, you can include the actual use case for reusing a single notification somewhere in the description (that is - the why and the benefits).

PR #2476, opened 2024-10-29
docs(common): Migrate to docs-builder

ntacheva:
Closes https://github.com/telerik/blazor/issues/9914


knowledge-base/splitter-save-and-load-state.md
```diff
@@ -95,7 +95,7 @@ The example below demonstrates the described approach.
     }
 }
 ````
-````Service
+````C# Service
```
dimodi, line 98: Should we use `C#` or `CS` ? If I am not mistaken, `cs` is the commonly used language name.

components/chart/multiple-axes.md
```diff
@@ -21,7 +21,7 @@ This article contains the following sections:
 	* [Numerical Charts](#numerical-charts)
 * [Choose Axis Position](#choose-axis-position)
 * [Examples](#examples)
-	* [Categorical Chart - Value Axes Examples](#categorical-chart---value-axes-examples)
+	* [Categorical Chart - Value Axes Examples](#categorical-chart-value-axes-examples)
```
dimodi, line 24: It looks like the algorithm for navigating to a heading has changed and some characters are ignored. Is there more information about this?

common-features/cdn.md
```diff
@@ -41,7 +41,7 @@ Here are a few examples:
 
 <div class="skip-repl"></div>
 
-````CHTML
+````RAZOR
```
dimodi, line 44: Do we need to use CAPS LOCK for the language name? It's more convenient without it.

components/chart/multiple-axes.md
ntacheva: Yes, the algorithm has been changed for some characters - some are ignored but some are additionally included (e.g. https://github.com/telerik/blazor-docs/pull/2476/files#diff-57dc4953244663e3f5151779ab49f1249b11d42b2d045cea757fe7a4c7673aadR55 - "Using a NuGet.Config file" where the dot is included). Do you mean that you'd like to get more information on the reasoning for this? Or, maybe some specific guidelines for the changes?

knowledge-base/splitter-save-and-load-state.md
ntacheva: I think we (with @radkostanev and @xristianstefanov) discussed that the `C#` is actually the more common one. I am not able to find some statistics that shows the opposite, though. Have you come across any?

common-features/cdn.md
ntacheva: I don't think we've discussed lower casing here. I tested `Razor` and it also seems to be working. However, I'd prefer to keep the capitalization `RAZOR` for consistency with the other languages. Plus, it is visualized like so regardless of how it is written in the documentation:
image

I'm not sure if this can be changed on our side. If som do you think we should request it?

components/chart/multiple-axes.md
dimodi: The goal of the question was to ensure we have a resource for future docs changes.

dimodi (approved):

common-features/cdn.md
ntacheva: TBD on the meeting tomorrow.

PR #2480, opened 2024-10-31
docs(Editor): Document built-in toolsets

dimodi:


components/editor/toolbar.md
```diff
@@ -33,22 +33,64 @@ To control the collection of buttons and commands available to the user, you pro
...
+>caption Use all built-in Editor tools
 
 ````CSHTML
 @using Telerik.Blazor.Components.Editor
 
 <TelerikEditor Tools="@EditorToolSets.All">
 </TelerikEditor>
+
+@code {
+    private string EditorValue { get; set; } = string.Empty;
```
Tsvetomir-Hr, line 52: Is this property required for the example?

Tsvetomir-Hr (approved):

components/editor/toolbar.md
dimodi: Yes, it's good to not provide invalid runnable configurations. I should add `@bind-Value`. Thanks!

PR #2550, opened 2024-11-21
chore(grid): clarify the required DotNetObjectReference type

ntacheva:
And code style fixes.


dimodi (approved):
We can also add this note before the example for better visibility. Otherwise the customer may hit the error first.

PR #2572, opened 2024-11-26
chore(common): update package source mapping

ntacheva:


yordan-mitev (approved):

dimodi (approved):

dimodi:
A possible improvement is to provide a complete `NuGet.Config` file here, because readers sometimes forget to match the Telerik feed name in the source settings and in the mapping setting.s

PR #2679, opened 2025-01-06
kb(grid): Fix percent example and polish explanation

dimodi:


knowledge-base/grid-adjust-height-with-browser.md
```diff
@@ -60,13 +60,19 @@ The following sections demonstrate different ways to make a Grid resize vertical
 
 Use percentage heights to define dimensions as a portion of the parent element.
 
-When setting a `height` in percent, keep in mind the following rule: an element with a percentage `height` requires its parent to have an explicit height. The rule applies recursively until a fixed height is reached, or until the `<html>` element is reached.
+When setting a `height` in percent, keep in mind the following rule: an element with a percentage `height` requires its parent to have an explicit height in any absolute or relative CSS unit. The rule applies recursively until a fixed height is reached (in absolute units), or until the `<html>` element is reached.
```
dimodi, line 63:
```suggestion
When setting a `height` in percent, keep in mind the following rule: an element with a percentage `height` requires its parent to have an explicit height in any absolute or relative CSS unit. The rule applies recursively until a fixed (absolute) height is reached, or until the `<html>` element is reached.
```

ntacheva (approved):

PR #2684, opened 2025-01-08
docs(grid): Fix misleading statements

dimodi:


xristianstefanov (approved):

ikoevska (commented):
Writing using "we" and "our" should be avoided as per the [style guide](https://docs.telerik.com/style-guide/cheat-sheet#basic-things-to-avoid).

PR #2731, opened 2025-01-24
Migrate API links to slugs

dimodi:


ikoevska (approved):
The syntax seems to be the correct one (looking only at the source) and everything builds both for dev and production without any issue (broken slugs are a bloker, so having a successful build is a good sign).

However, even though I was able to build both the docs and the API ref locally, I haven't been able to run the site successfully. Hence, I wasn't able to verify that everything works. Keep in mind that you might need to do a sanity check when deployed on staging.

PR #2855, opened 2025-03-21
kb(Dropdowns): Add KB for disabled text wrapping

dimodi:


xristianstefanov (approved):

knowledge-base/dropdowns-disable-long-text-wrap.md
```diff
@@ -0,0 +1,113 @@
...
+            </td>
+        </tr>
+    </tbody>
+</table>
+
+## Description
+
+This KB article deals with the following scenarios:
+
+1. The items in a Telerik dropdown component are long and wrap to multiple lines. The items become very high. How to avoid and disable this behavior?
```
ikoevska, line 32: These should not be an ordered list, since they are two separate scenarios. I recommend going with simple bullets instead (unordered list).

ikoevska (commented):
Reviewed.

PR #2986, opened 2025-05-21
chore(PanelBar): update field bindings docs

ntacheva:
Closes https://github.com/telerik/blazor/issues/5075

### DIFF
--- a/components/panelbar/data-binding/overview.md
+++ b/components/panelbar/data-binding/overview.md
@@ -66,7 +66,7 @@ Each `PanelBarBinding` tag exposes the following properties that refer to item p
 
 * ItemsField => Items
 
-* Level - this is used for defining [custom field bindings](#custom-field-bindings) or [different bindings for different levels](#multiple-level-bindings). If no level is set, the bindings are taken as default for any level that does not have explicit settings. You should have one `TelerikPanelBarBinding` without a level to set the default bindings.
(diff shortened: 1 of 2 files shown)


components/panelbar/data-binding/overview.md
```diff
@@ -66,7 +66,7 @@ Each `PanelBarBinding` tag exposes the following properties that refer to item p
 
 * ItemsField => Items
 
-* Level - this is used for defining [different bindings for different levels](#multiple-level-bindings). If no level is set, the bindings are taken as default for any level that does not have explicit settings. You should have one `TelerikPanelBarBinding` without a level.
+* Level - this is used for defining [custom field bindings](#custom-field-bindings) or [different bindings for different levels](#multiple-level-bindings). If no level is set, the bindings are taken as default for any level that does not have explicit settings. You should have one `TelerikPanelBarBinding` without a level to set the default bindings.
```
ikoevska, line 69:
```suggestion
* Level&mdash;this is used for defining [custom field bindings](#custom-field-bindings) or [different bindings for different levels](#multiple-level-bindings). If no level is set, the bindings are taken as default for any level that does not have explicit settings. You should have one `TelerikPanelBarBinding` without a level to set the default bindings.
```

ikoevska: Should is not the best choice of a modal verb. See [here ](https://docs.telerik.com/style-guide/basic-rules-and-guidelines#use-properly-should-could-must-have-to-need-to) for more information.

ikoevska (approved):

PR #3100, opened 2025-07-14
docs(Grid): add highlighting docs

Tsvetomir-Hr:
closes: https://github.com/telerik/blazor/issues/11690


ikoevska (approved):
Think some more about your cross-linking strategy. Are there other locations from which you can link TO the new article?

PR #3447, opened 2026-01-15
docs: address weekly feedback

IvanDanchev:


dimodi (commented):
Commenting one line out can cause confusion. Consider these options:

* Leave the comment, but clarify the reason in the code itself and in the text before the example.
* Remove the comment and make the whole example non-runnable, because when it's runnable, it cannot fully demonstrate the discussed scenario.

In either case, the example looks complex for its simple goal. Consider adding one more example that uses the approach from our project template: use `ExpandedChanged` and do not update the `Expanded` parameter value.

knowledge-base/drawer-prevent-collapse.md
```diff
@@ -27,79 +27,50 @@ I would like to prevent the Drawer from collapsing when an item from the navigat
...
-                        {
-                            <span class="k-item-text">@item.Text</span>
-                        }
-                    </li>
-                }
-            </ul>
-        </div>
-    </Template>
+               MiniMode="true">
     <DrawerContent>
```
dimodi, line 42: Consider rendering the item `Text` here, otherwise the Drawer doesn't appear to do anything on item change.

dimodi (approved):

PR #3650, opened 2026-04-26
docs: add two way binding and visible parameter info

IvanDanchev:
Closes https://github.com/telerik/blazor/issues/13691
Closes https://progresssoftware.atlassian.net/browse/BLZ-2492

### DIFF
--- a/components/dockmanager/pane-types.md
+++ b/components/dockmanager/pane-types.md
@@ -38,7 +38,7 @@ Organizes panes in a [Splitter-like](slug:splitter-overview) manner, allowing th
 
 ## Restore Closed Panes
 
-Restore closed panes by tracking their visibility state in a collection. Use two-way binding for the `Size` parameter, to persist the pane size.  
+To restore closed panes, track their visibility state in a collection. Use two-way binding for the `Size` parameter to persist the pane size.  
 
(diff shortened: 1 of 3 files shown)


components/dockmanager/pane-types.md
```diff
@@ -36,6 +36,95 @@ Organizes panes in a [Splitter-like](slug:splitter-overview) manner, allowing th
 * It can contain `<DockManagerTabGroupPane>`, `<DockManagerContentPane>`, and other `<DockManagerSplitPane>` tags as children. 
 * Only this pane type can be declared as a direct child of the `<DockManagerFloatingPanes>` tag.
 
+## Restore Closed Panes
+
+Restore closed panes by tracking their visibility state in a collection. Use two-way binding for the `Size` parameter, to persist the pane size.
```
yordan-mitev: - We still write for humans, who scan through the article looking for signals like to+infinitive, to identify the goal much faster than with the imperative mood as an opener.
- Consider tweaking the Copilot skill or instruction, so this specific is also respected when creating new content.
```suggestion
To restore closed panes, track their visibility state in a collection. Use two-way binding for the `Size` parameter to persist the pane size.  
```

dimodi (approved):

PR #3676, opened 2026-05-14
chore(Notification): rename OnClose event to OnHide

georgedautov:


IvanDanchev (approved):

georgedautov:
Closed this because the event will actually remain OnClose, because of the `Closeable` and `CloseAfter` parameters of the NotificationModel.

PR #3824, opened 2026-07-29
chore(workflow): added vpat generation workflow

MilenaCh:

### DIFF
--- a/.github/workflows/generate-vpat.yml
+++ b/.github/workflows/generate-vpat.yml
@@ -1,0 +1,83 @@
+name: Generate VPAT
+
+on:
+  push:
+    branches:
+      - 'vpat'
+      - 'vpat/**'
+  workflow_dispatch:
+    inputs:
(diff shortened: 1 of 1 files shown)


.github/workflows/generate-vpat.yml
```diff
@@ -0,0 +1,83 @@
...
+jobs:
+  generate:
+    uses: telerik/vpat-generator/.github/workflows/generate-vpat.yml@develop
+    permissions:
+      id-token: write
+      contents: read
+    with:
+      product: blazor
+      vpat_ref: ${{ github.event.inputs.vpat_ref || 'develop' }}
+      product_version: ${{ github.event.inputs.version || '14.1.0' }}
```
dimodi, line 27: If `inputs.version` is required, when do we expect it to be missing? Can we fail the workflow if the version is not available, rather than hard-code a number that must be maintained in the future?

